PR #138, opened 2019-01-29
Specify dependencies for pip install

### LINKED ISSUE #135: Make setup.py require dash
Really just a suggestion, but I believe it would be nice to have setup.py require dash by default:
https://github.com/plotly/dash-bio/blob/bd5ce53c4529ce3e0c783cb61eb6eab7b298dd93/setup.py#L11-L20

That way, when the user installs dash-bio, they simply need to create a venv and run:
```
pip install dash-bio
```
with dash, dash-html-components, dash_renderer being installed as well.

It could be implemented along the line of:

```python
 setup( 
     name=package_name, 
     version=package["version"], 
     author=package['author'], 
     packages=[package_name, '{}/utils'.format(package_name), '{}/component_factory'.format(package_name)], 
     include_package_data=True, 
     license=package['license'], 
     description=package['description'] if 'description' in package else package_name, 
     install_requires=[
         'dash', 'dash-html-components', 'dash_renderer'
     ] 
 ) 
```

mkcor:
Closes #135 (see issue for discussion).


setup.py
```diff
@@ -16,5 +16,8 @@
     include_package_data=True,
     license=package['license'],
     description=package['description'] if 'description' in package else package_name,
-    install_requires=[]
+    install_requires=[
```
shammamah: We might want to specify the version of `dash` we use here! 

Also, where do we need `dash-html-components`? I don't think it's used anywhere in `dash-bio` itself, just in the demo apps. 

Might also be a good idea to go through the pure Python components (clustergram/manhattan plot/volcano plot) to see what their requirements are, and cross-reference with `requirements.txt` to get the exact versions. What do you think?

xhlulu:
I think Alex mentioned a specific version of dash that already requires all the other dependent libraries (html, dcc, etc.), is that version released yet?

alexcjohnson:
> I think Alex mentioned a specific version of dash that already requires all the other dependent libraries (html, dcc, etc.), is that version released yet?

No, that's still TODO https://github.com/plotly/dash-core/issues/24

I'd include `dash-html-components` explicitly for now, so it can be omitted from the readme and `dash-bio` is a complete app-development framework on its own. Once ^^ is done we can leave the readme as is and just pare down `setup.py`

setup.py
shammamah: I think we should keep those separate, since we're just providing a library of components and not an app-building library.

>For the dash version, what version should be the minimum?

Let's go with `0.36.0`, since that represents a big change (removal of events).

shammamah: Oops, just saw @alexcjohnson's reply. Let's do that instead!

mkcor: @shammamah So you're suggesting specifying 'dash>=0.36.0'? I guess, at this point, we might as well wait for the release @alexcjohnson mentioned, and then we'll specify 'dash>=[that release version]' and remove 'dash-html-components'...

Regarding cross-referencing dependencies in `requirements.txt`, I'm not in favour of it; I have reread https://packaging.python.org/discussions/install-requires-vs-requirements/, I have [quoted](https://github.com/plotly/dash-bio/issues/135#issuecomment-458467753) it before, and `requirements.txt` fit a different purpose.

Issue #133, opened 2019-01-25
Dash Bio apps need to be moved to dash-bio.plotly.host


shammamah:
In accordance with https://github.com/plotly/streambed/issues/12263, we are making our `index.py` app obsolete. The applications we currently have are not true standalone apps, since they just modify the layout on the main `index.py` app. Right now, `index.py` lives on that server (https://dash-gallery.plotly.host/dash-bio) and its portal (https://dash-bio.plotly.host/Portal/)

What we need to do: 
- [x] Publish the dash bio package to test.pypi.org 
- [x] Separate the requirements for each app from our `requirements.txt` file, add in the `dash_bio` dependency
- [ ] Set up a file structure within the repository that will house all of the applications. I'm thinking just a folder called `demo_apps` that contains a folder for each app 
- [ ] Copy each app's relevant `sample_data` files into its folder
- [ ] Create all the necessary files (looks like we need a `config.py` and `runtime.txt`) for each app in its folder
- [ ] Once the apps are on the portal, tag them (this looks like a really cool feature we should take advantage of!) with things like "genomics", "visualizations", etc. 

Let's get this done before I fly out to Boston for training with Amgen (13 Feb)! They're a huge pharmaceutical company and it would be awesome to show off Dash Bio and sell them on DDS simultaneously :) let's get that 🍞

shammamah:
@mkcor @Bachibouzouk We can easily convert each `app_[]` file to a standalone app by adding the following lines: 

```
app = dash.Dash()
app.scripts.config.serve_locally = True
server = app.server
app.layout = layout() # assign layout 
app.callbacks() # register all of the callbacks 
``` 
somewhere in the beginning of the file and 

```
if __name__ == '__main__':
    app.run_server(debug=True)
``` 

at the end. Will try this out with one of our applications and confirm :)

Bachibouzouk:
@shammamah : I did 
```
import dash


def run_standalone_app(layout, callbacks, port=8050, debug=True):

    app = dash.Dash()
    app.scripts.config.serve_locally = True
    app.config["suppress_callback_exceptions"] = True

    server = app.server

    app.layout = layout()  # assign layout
    callbacks(app)  # register all of the callbacks
    app.run_server(debug=debug, port=port)

```
in a `app_standalone.py` file in `tests/dashbio_demos/utils`, then 

```
if __name__ == '__main__':
    from tests.dashbio_demos.utils.app_standalone import run_standalone_app
[...]

PR #144, opened 2019-01-31
Let demo apps run as standalone apps as well.

mkcor:
This PR partly addresses #133.

Do you think that, just for this PR, I should revert b270c73b0b282210a3689d0c7ca70f1d5bed53d9 so that CI tests pass, and get this step merged in?

Note that dc1c769e16032530350ce7a2630059be0f975fae introduced some '^M' line-ending characters (as hinted in a previous [comment](https://github.com/plotly/dash-bio/issues/133#issuecomment-458508075) of mine). It doesn't show on GitHub, but this was the output of `git diff` at the command line:

line_ending_char

PR #149, opened 2019-02-04
Let example/demo apps run as standalone apps.

mkcor:
Lays the ground for #133.

With this PR, it is still possible to run `python index.py` from the root of the repo, but now also possible to run `python tests/dashbio_demos/app_ideogram.py`.

What's left to do:

- [x]  I have run `pip install -i https://test.pypi.org/simple/ dash-bio` in my (activated) virtualenv, but I'm not sure about the syntax to add it to `tests/requirements.txt`...

- [x]  Read again something like https://chrisyeh96.github.io/2017/08/08/definitive-guide-python-imports.html because imports are of different kinds in these changesets and I'm confused...

- [ ]  Add the same `if __name__ == '__main__':` block to all other example apps.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ biopython
 colour==0.1.5
 cython>=0.19
 dash==0.35.2
-dash_bio==0.0.1
+dash-bio==0.0.2
 dash-core-components==0.36.1
 dash-html-components==0.13.2
 dash-renderer==0.15.0
--- a/tests/dashbio_demos/utils/app_wrapper.py
+++ b/tests/dashbio_demos/utils/app_wrapper.py
@@ -5,8 +5,9 @@ import base64
 
 def app_page_layout(page_layout,
                     app_title="Dash Bio App",
-                    app_github_url="",
+                    app_name="",
                     light_logo=True,
+                    standalone=False,
                     bg_color="#506784",
                     font_color="#F3F6FA"):
     return html.Div(
@@ -47,8 +48,11 @@ def app_page_layout(page_layout,
                                 )
                             )
                         ],
-                        href="http://github.com/plotly/dash-bio/"
-                             "blob/master/tests/dashbio_demos/app_{}.py".format(app_github_url)
+                        href="/" if standalone else
+                        "http://github.com/plotly/dash-bio/"
+                        "blob/master/tests/dashbio_demos/app_{}.py".format(
+                            app_name
+                        )
                     )
                 ],
                 style={
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,3 +1,5 @@
+--extra-index-url https://test.pypi.org/simple/
+
 # Packages needed to run the tests.
 # Switch into a virtual environment
 # pip install -r requirements.txt
@@ -6,6 +8,7 @@ chromedriver-binary==2.45.0
 colour==0.1.5
 cython>=0.19
 dash==0.35.2
+dash-bio==0.0.2
 dash-core-components==0.36.1
 dash-html-components==0.13.2
 dash-renderer==0.15.0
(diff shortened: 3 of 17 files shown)


mkcor:
So, the fail on CircleCI is due to a pylint error already documented by @Bachibouzouk in this comment: https://github.com/plotly/dash-bio/pull/144#pullrequestreview-198624734

If I add the `.` in front the relative(?) import, I cannot run `python tests/dashbio_demos/app_ideogram.py` anymore:
```bash
Traceback (most recent call last):
  File "tests/dashbio_demos/app_ideogram.py", line 860, in <module>
    from .utils.app_standalone import run_standalone_app
ModuleNotFoundError: No module named '__main__.utils'; '__main__' is not a package
```
which is why I need to read again the Definitive Guide to Python Imports (or get some sleep).

shammamah:
@mkcor You just need to add 
```
--extra-index-url https://testpypi.pypi.org/simple/
```
to the beginning of `requirements.txt`. (Reference: https://stackoverflow.com/questions/29289695/installing-packages-from-multiple-servers-from-one-or-more-requirements-file)

mkcor:
> @mkcor You just need to add
> 
> ```
> --extra-index-url https://testpypi.pypi.org/simple/
> ```

@shammamah Oh, just like you did at the root-level `requirements.txt`! I have taken this opportunity to fix up the `dash-bio` dependency (not to be confused with the `dash_bio` import, as we were discussing!) and bump up its version. Is that okay?

mkcor:
After studying relative/absolute module/package imports for the n-th time, I'm reminded that there is more than one way to do them... So I went for tweaking `.pylintrc` instead! I cherry-picked the corresponding commit from the previous, work-in-progress branch.

requirements.txt
```diff
@@ -6,7 +6,7 @@ biopython
 colour==0.1.5
 cython>=0.19
 dash==0.35.2
-dash_bio==0.0.1
+dash-bio==0.0.2
```
shammamah:
```suggestion
dash-bio==0.0.3
```

tests/requirements.txt
```diff
@@ -6,6 +8,7 @@ chromedriver-binary==2.45.0
 colour==0.1.5
 cython>=0.19
 dash==0.35.2
+dash-bio==0.0.2
```
shammamah:
```suggestion
dash-bio==0.0.3
```

tests/dashbio_demos/utils/app_wrapper.py
```diff
@@ -5,8 +5,9 @@
 
 def app_page_layout(page_layout,
                     app_title="Dash Bio App",
-                    app_github_url="",
```
Bachibouzouk: @mkcor  if this argument is renamed here, it should be refactored in `index.py` as well, this is what make the Integration tests fail on CircleCI

mkcor:
@Bachibouzouk did you see my comment https://github.com/plotly/dash-bio/pull/149#issuecomment-460465735 and the related commit f0b6f5d? This comment updated the earlier one about CircleCI tests failing (where I tagged you), and this commit made the CI tests pass. The CI tests were passing until I bumped the version number for dash-bio (commit 8105bc9).

mkcor:
@shammamah Please write short enough commit messages (and without typos).

tests/dashbio_demos/utils/app_wrapper.py
shammamah: Fixed in https://github.com/plotly/dash-bio/pull/149/commits/531a8885586fa794eb97be6394aa9f8a626d09e4.

mkcor:
Thank you so much, @shammamah! Can you approve (even if you authored the latest commits...)? I'll merge since I originally submitted the PR.

PR #190, opened 2019-02-19
Customize the bool function in test_common_features

Bachibouzouk:
## Description of changes 

This function will return `False` if its argument is `'false'`, `'False'` or `'0'` whereas builtin `bool()` would return `True`


tests/test_common_features.py
```diff
@@ -15,10 +15,25 @@
 COMPONENT_PYTHON_BASE = 'python'
 COMPONENT_REACT_BASE = 'react'
 
+
+def bool_converter(s):
+    """Convert variable to boolean
```
mkcor:
```suggestion
    """Return the same as built-in function bool() except for arguments which are
    string representations of a boolean value.
```

tests/test_common_features.py
```diff
@@ -15,10 +15,25 @@
...
+
+    :param s: a variable
+    :return: the return of the builtin bool() function except if the variable is equal to a str
```
mkcor:
```suggestion
    :return: True or False
```

tests/test_common_features.py
```diff
@@ -15,10 +15,26 @@
...
 COMPONENT_REACT_BASE = 'react'
 
+
+def bool_converter(s):
+    """Return the same as built-in function bool() except for arguments which are
+    string representations of a boolean value.
+
+    :param s: a variable
+    :return: True or False
+    representation of the boolean value
```
mkcor: @Bachibouzouk You want to remove this line.
I wasn't sure how to do it in suggestion mode, because blank line is not the same as no line...

mkcor:
Good catch, @Bachibouzouk! Thanks for handling it.

tests/test_common_features.py
```diff
@@ -15,10 +15,26 @@
 COMPONENT_PYTHON_BASE = 'python'
 COMPONENT_REACT_BASE = 'react'
 
+
+def bool_converter(s):
```
shammamah: We can instead check if the input parameter is equal to a valid boolean value (and if not, throw an `Exception`) -- there are many ways that someone can mistype `True` or `False` :)

tests/test_common_features.py
```diff
@@ -15,10 +15,34 @@
...
+    """
+    answer = bool(s)
+    if isinstance(s, str):
+        if s in ('False', 'false', '0'):
+            answer = False
+        elif s in ('True', 'true', '1'):
+            answer = True
+        else:
+            raise ValueError(
+                'Expected one of {}, recieved : {}'.format(
```
mkcor:
```suggestion
                'Expected one of {}, received : {}'.format(
```

Issue #103, opened 2019-01-15
Write tests for all components/apps 


shammamah:
Now we have integration testing set up, we need a common test suite that is used for all common features and files specific to each component's features. 

@Bachibouzouk has already started on his (#102) for Volcano plot. 

We still need to finish tests for: 
- [x] general (to be used for all components) (#142, #146, #173)
- [ ] Alignment Chart 
- [ ] Circos 
- [x] Ideogram (#182)
- [ ] Molecule3D 
- [x] NeedlePlot (#146) 
- [ ] SequenceViewer 
- [ ] Speck 
- [ ] Clustergram 
- [x] Manhattan plot (#142)
- [x] Volcano plot (#142, #102)
- [ ] Oncoprint (#191, #194)

shammamah:
@mkcor @Bachibouzouk I'm thinking we can split these up. 

@shammamah 
* Speck 
* Sequence viewer 
* Clustergram 

@Bachibouzouk 
* Alignment chart
* Circos 
* Manhattan plot 
* Needle plot

@mkcor 
* Molecule3D
* Oncoprint 
 * Ideogram 

Let me know if you want to switch anything around!

Bachibouzouk:
@shammamah, @mkcor : I made tests for testing the app works correctly as well as tests for testing the passing of props to the component, ping me if you want to chat about the way I carry out the tests.

mkcor:
@Bachibouzouk Wonderful! Do you mean it would be more fruitful to have this chat *before* I review #142 and #143 ?

Bachibouzouk:
Useful command to run only one test `pytest --driver Chrome tests/<test file>::<test function within the file>`

Bachibouzouk:
Useful to check if the react component changed after modifying a prop value
https://selenium-python.readthedocs.io/locating-elements.html#locating-elements-by-class-name

mkcor:
As per today's discussion, we want to make sure we are defensive in our demo apps "and the most common way to do it is to add assertions to our code" (https://swcarpentry.github.io/python-novice-inflammation/08-defensive/index.html).
/cc @shammamah @Bachibouzouk

PR #194, opened 2019-02-22
Write tests for Layout props (OncoPrint component).

mkcor:
Fits in #103.

## About 
- This is improving our test coverage.

## Description of changes 
I'm going through the list of props for component OncoPrint and decided to complete tests for those under "Layout" in the demo app (showoverview, showlegend, padding: https://dash-bio.plotly.host/dash-oncoprint/ -- which should soon become https://dash-bio.plotly.host/dash-onco-print/?), although what falls under "Layout" is described [otherwise](https://github.com/plotly/dash-bio/blob/e16f5405bc052d70fbb1c2da52969c94ab55e709/src/lib/components/OncoPrint.react.js#L131) in the source code (range, showlegend, showoverview)...?

The build fail for 0d8870f is strange (only for Python 3.7), as if the doubling of chromosomes didn't always succeed:
```
        # assert doubling of the 22 chromosomes + X and Y chromosomes
        num_chromosoms = len(driver.find_elements_by_class_name('chromosome'))
>       assert num_chromosoms == 46
E       assert 24 == 46
```

## Before merging


tests/test_onco_print.py
```diff
@@ -73,7 +74,29 @@ def oncoprint_props_callback(
...
 
-# below are tests for changing the props of the React component
+# Tests for layout props (under "Customize" tab)
+def test_showoverview(dash_threaded):
+    """Test the overview display."""
+    def assert_callback(prop_value, nclicks, input_value):
+        answer = ''
+        if nclicks is not None:
+            if PROP_TYPES['bool'](input_value) == prop_value:
+                answer = 'PASSED'
```
shammamah: In [the `test_skeleton` file](https://github.com/plotly/dash-bio/blob/e16f5405bc052d70fbb1c2da52969c94ab55e709/tests/test_skeleton.py#L22), we have this: 

```
# pass/fail strings
PASS = 'PASSED'
FAIL = 'FAILED'
```

Do you think it's a good idea to include this? @Bachibouzouk is an advocate of "hard-coding" strings like this to avoid errors that could pop up if someone makes a typo.

mkcor:
From now on, I should probably define a function similar to @Bachibouzouk's [`generate_subprop_test_needle()`](https://github.com/plotly/dash-bio/blob/e16f5405bc052d70fbb1c2da52969c94ab55e709/tests/test_needle_plot.py#L47) in order to avoid copying-pasting a lot of almost equal code... But I wonder why it's calling `test_common_features`'s `generate_subprop_test()`](https://github.com/plotly/dash-bio/blob/e16f5405bc052d70fbb1c2da52969c94ab55e709/tests/test_common_features.py#L270), which is defined (only) for prop type 'dict'...

shammamah:
>although what falls under "Layout" is described [otherwise](https://github.com/plotly/dash-bio/blob/e16f5405bc052d70fbb1c2da52969c94ab55e709/src/lib/components/OncoPrint.react.js#L131) in the source code (range, showlegend, showoverview)...?

Just took a look and that doesn't make much sense to me, either. Let's just define all "layout"-related props as ones that have to do with changing the appearance of the component somehow. I think all of the props in the `defaultProps` can classify as "layout" types.

tests/test_onco_print.py
mkcor: Good catch. I kept these variables in `test_onco_print.py` and I'm not making use of them, so it's not consistent... I would be in favour of using the `PASS` variable.

Bachibouzouk:
> From now on, I should probably define a function similar to @Bachibouzouk's [`generate_subprop_test_needle()`](https://github.com/plotly/dash-bio/blob/e16f5405bc052d70fbb1c2da52969c94ab55e709/tests/test_needle_plot.py#L47) in order to avoid copying-pasting a lot of almost equal code... But I wonder why it's calling `test_common_features`'s [`generate_subprop_test()`](https://github.com/plotly/dash-bio/blob/e16f5405bc052d70fbb1c2da52969c94ab55e709/tests/test_common_features.py#L270), which is defined (only) for prop type 'dict'...

@mkcor : a subprop is contained in a prop which is a dict (`PropTypes.shape()` in react). OncoPrint has no subprops. I would recommand the use of `template_test_component` instead

tests/test_onco_print.py
```diff
@@ -94,3 +117,25 @@ def assert_callback(prop_value, nclicks, input_value):
...
         data=TEST_DATA
     )
+
+
+def test_padding(dash_threaded):
+    """Test the updating of the padding value."""
+    def assert_callback(prop_value, nclicks, input_value):
+        answer = ''
+        if nclicks is not None:
+            if float(input_value) == prop_value:
```
mkcor: @shammamah @Bachibouzouk here I hesitated with this:
```suggestion
            if PROP_TYPES['float'](input_value) == prop_value:
```
allowing the flexibility (possible future change?) of this [definition](https://github.com/plotly/dash-bio/blob/e16f5405bc052d70fbb1c2da52969c94ab55e709/tests/test_common_features.py#L44). What do you think?

shammamah: This is a great idea! We had to change the `bool` function, so we should use this kind of notation for all of the type conversions from strings.

Bachibouzouk: @ mkcor that is exactly the spirit here :) :+1:

mkcor:
@Bachibouzouk @shammamah I must say I don't know how to check the actual updating of the padding value... It's not just an element you can find in the webpage...

mkcor:
Regarding `showoverview`, I'm currently blocked by this issue: #214

tests/test_onco_print.py
```diff
@@ -94,3 +120,38 @@ def assert_callback(prop_value, nclicks, input_value):
         component_base=COMPONENT_REACT_BASE,
         data=TEST_DATA
     )
+
+    driver = dash_threaded.driver
+    # assert there is a legend (bar)
+    legend = driver.find_elements_by_class_name('legendbar')
```
Bachibouzouk:
```suggestion
    legend = wait_for_elements_by_css_selector(driver, '.legendbar')
```
In order to avoid test failure if the component take time to update

tests/test_onco_print.py
```diff
@@ -94,3 +120,38 @@ def assert_callback(prop_value, nclicks, input_value):
...
+    assert len(legend) != 0
+
+    # trigger change of the component prop
+    btn = wait_for_element_by_css_selector(driver, '#test-{}-btn'.format(APP_NAME))
+    btn.click()
+
+    # assert there is no more legend (bar)
+    legend = driver.find_elements_by_class_name('legendbar')
```
Bachibouzouk:
```suggestion
    legend = wait_for_elements_by_css_selector(driver, '.legendbar')
```

tests/test_onco_print.py
```diff
@@ -1,7 +1,10 @@
 import os
+
+from pytest_dash.wait_for import wait_for_element_by_css_selector
```
Bachibouzouk:
```suggestion
from pytest_dash.wait_for import wait_for_element_by_css_selector, wait_for_elements_by_css_selector
```

tests/test_onco_print.py
```diff
@@ -94,3 +120,38 @@ def assert_callback(prop_value, nclicks, input_value):
...
+        dash_threaded,
+        APP_NAME,
+        assert_callback,
+        oncoprint_props_callback,
+        'padding',
+        '0.08',
+        prop_type='float',
+        component_base=COMPONENT_REACT_BASE,
+        data=TEST_DATA
+    )
```
Bachibouzouk: Isn't the test missing here?

tests/test_onco_print.py
```diff
@@ -73,17 +77,39 @@ def oncoprint_props_callback(
...
+        dash_threaded,
+        APP_NAME,
+        assert_callback,
+        oncoprint_props_callback,
+        'showoverview',
+        'False',
+        prop_type='bool',
+        component_base=COMPONENT_REACT_BASE,
+        data=TEST_DATA
+    )
```
Bachibouzouk: Isn't the test missing here?

tests/test_onco_print.py
mkcor: ... if the *prop* took time to update, right?

Oops, I didn't notice the difference between `wait_for_element_by_css_selector()` and `wait_for_elements_by_css_selector()`!!

tests/test_onco_print.py
mkcor: Reading function definition for [`wait_for_elements_by_css_selector()`](https://github.com/T4rk1n/pytest-dash/blob/2ea221fad6913f168693212986b28b576fca091c/pytest_dash/wait_for.py#L44). I wish it was called something *more* different, such as "wait_for_all_elements" but I understand the function name is already very long ;) @T4rk1n

tests/test_onco_print.py
mkcor: Aww I'll stack these imports over multiple lines, because it's way too confusing...

tests/test_onco_print.py
mkcor: Oh, I think I just inferred the meaning of "Add suggestion to batch"! We can probably include several suggestions into one commit, which I should have done here, so it makes sense conceptually *and* CI tests pass!
/cc @shammamah

mkcor:
@Bachibouzouk Ironically enough, now that I've applied the suggestions to avoid timeout errors, we're getting timeout errors!! https://circleci.com/gh/plotly/dash-bio/1326

mkcor:
@shammamah This PR is not as complete as I originally intended it to be, but it still brings some improvement. In this light, can you please review and approve it? I don't want us to end up with multiple open PRs on long-lived feature branches again... I'll address #214 in a subsequent PR. Thanks.

PR #202, opened 2019-02-25
Reorder function arguments in boilerplate code.

mkcor:
## About
As per our February 19 discussion, we might as well order arguments of the `assert_callback()` function as nclicks, prop name, ... to match the order suggested in the `COMPONENTNAME_props_callback()` function, declared before.

## Description of changes
This change is not functional, since this code never runs. It's just easier on the brain to read "nclicks, prop name, ..." and then again "nclicks, prop name, ..." (rather than "prop name, niclicks, ..." on the second time).

## Before merging

### DIFF
--- a/tests/test_skeleton.py
+++ b/tests/test_skeleton.py
@@ -77,9 +77,9 @@ def COMPONENTNAME_props_callback(
     """This function is the code of a callback which is triggered by
     the button on the simple app used in the test.
     :param nclicks (int): The n_clicks value of the button in the
-                          simple app
+                          simple app.
     :param prop_name (string): The name of the property that is to be
-                               modified
+                               modified.
     :param prop_value (string): The value that is to be assigned to the
                                 prop defined by prop_name.
     :prop_type (string): One of the predefined types in PROP_TYPES.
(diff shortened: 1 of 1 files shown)


tests/test_skeleton.py
```diff
@@ -110,16 +110,16 @@ def test_PROPNAME(dash_threaded):
...
     ):
         """Determine the pass/fail status of this test.
 
-        :param (string) component_PROPNAME: The value of PROPNAME for the
-                                            component after it is set
         :param (int) nclicks: The number of clicks on the button in the
                               simple test app (not used here)
+        :param (string) component_PROPNAME: The value of PROPNAME for the
+                                            component after it is set
         :param (string) input_PROPNAME: The value of PROPNAME that is sent
```
shammamah: Small nitpick - if this docstring is a full sentence with a period, should the other ones follow suit?

mkcor: Yes, I originally added the period when editing, then noticed that other param descriptions didn't end with a period, except for the last param... Will add periods everywhere then!

mkcor: Also, should we use the same param description for both `nclicks`?

"The n_clicks value of the button in the simple app." and "The n_clicks value of the button in the simple app (not used here)." instead of "The number of clicks on the button in the simple test app (not used here)."

shammamah: I think that's a good idea!

mkcor: Thanks, done! Meanwhile, I have also 'fixed' the order argument name/type.

shammamah:
💃 (Waiting for the CI checks to pass before I "officially" approve!)

mkcor:
> (Waiting for the CI checks to pass before I "officially" approve!)

For sure! Now they've passed.

mkcor:
Except I have to merge in master!

Issue #203, opened 2019-02-25
Errors in Ideogram component showing up in its demo app


Bachibouzouk:
The annotation tooltip does not display correctly as in https://eweitz.github.io/ideogram/annotations-tracks:

![screenshot from 2019-02-25 19-01-41](https://user-images.githubusercontent.com/4399407/53358200-f3521600-392f-11e9-959b-cd8e7a9abb70.png)

The `homology` prop seems to not work properly as in https://eweitz.github.io/ideogram/homology-basic:
![screenshot from 2019-02-25 19-02-02](https://user-images.githubusercontent.com/4399407/53358203-f4834300-392f-11e9-9670-bb1c1e750554.png)

PR #204, opened 2019-02-25
Fix ideogram homology

Bachibouzouk:
Addesses part of #203 
## About 
- [ ] This is a new component 
- [x] I am closing an issue 

## Description of changes 
Wrote three tests for the props `homology` and `fullChromosomeLabels` of `Ideogram` component. 

Modified the .react.js file so that the prop `fullChromosomeLabels` can be updated through a callback.

## Before merging


dash_bio/Ideogram.py
```diff
@@ -155,7 +155,9 @@ class Ideogram(Component):
 cytogenetically well-characterized organisms, e.g. human, beside chromosomes of
 less studied organisms, e.g. chimpanzee.
 - showNonNuclearChromosomes (boolean; optional): Whether to show non-nuclear chromosomes,
-e.g. for mitochondrial (MT) and chloroplast (CP) DNA."""
+e.g. for mitochondrial (MT) and chloroplast (CP) DNA.
+
+Available events: """
```
shammamah: This was generated using the wrong version of Dash -- please use `0.37.0` or higher!

tests/dashbio_demos/app_ideogram.py
```diff
@@ -802,12 +781,12 @@ def ideo_homology(
             stop_two_a
     ):
         return {
-            "chr_one": {
+            "chrOne": {
```
shammamah: Why aren't we using snake case here?

Bachibouzouk: Because the prop names in the `.react.js` file are in camel case

shammamah:
When you click on a chromosome it doesn't rotate and this error pops up: 
screen shot 2019-03-05 at 2 26 11 pm

Bachibouzouk:
> When you click on a chromosome it doesn't rotate and this error pops up:
@shammamah I had already reported it in #203 :). I went further and regenerated the component after locking the `package.json` version to the release [1.2.0](https://github.com/eweitz/ideogram/commit/4a8e526c73fe38b4c50d01f8a9ead537b981d5d5) of Ideogram and there the component rotates on click. But if I change it back to the latest release [1.5.0](https://github.com/eweitz/ideogram/releases/tag/v1.5.0), it shows this error. I propose to fix the version to [1.5.0](https://github.com/eweitz/ideogram/releases/tag/v1.5.0), because many ideogram tests now fail with [1.2.0](https://github.com/eweitz/ideogram/commit/4a8e526c73fe38b4c50d01f8a9ead537b981d5d5).

shammamah:
So the version on `master` is using 1.2.0? The Ideogram tests seem to pass on `master`.

Bachibouzouk:
> So the version on `master` is using 1.2.0? The Ideogram tests seem to pass on `master`.

The version on master is not using 1.2.0 as the app on https://dash-bio.plotly.host/dash-ideogram/ also has the rotation problem. The [commit number](https://github.com/eweitz/ideogram/commit/05cdc8c233e586bf97c38a8c215262cfd2d9ca49) in `package-lock.json` is in release 1.4.0, dating from late november 2018.

shammamah:
I see. Let's get the rotation issue fixed before we merge this one in.

Bachibouzouk:
> I see. Let's get the rotation issue fixed before we merge this one in.

@shammamah  : to make the PR more concise, wouldn't it better to tackle this issue in a subsequent PR, together with a test for the `rotatable` prop?

shammamah:
If it will be a short fix, sure. I just would prefer for everything on `master` to work properly. I'll approve this one!

PR #206, opened 2019-02-25
Remove unicode string.

### LINKED ISSUE #205: Circos unicode character leads to errors with Python 2.7
While writing documentation for the Dash Bio components, I ran into an issue that involved this specific unicode character: https://www.fileformat.info/info/unicode/char/27a4/index.htm

This is only used in the Circos file (e.g., [here](https://github.com/plotly/dash-bio/blob/0214fb0d9335b387639d3a1b6e8fd8906cb065ed/src/lib/components/Circos.react.js#L145)), and doesn't appear to be super important. 

I have since re-built `dash-bio` temporarily as an [`rc` release](https://test.pypi.org/project/dash-bio/0.0.5rc1/) after removing the character, and the docs (as well as Python 2.7) are cooperating with it, so all we really need to do is change that single character.

shammamah:
Closes #205 

## About 
- [ ] This is a new component 
- [x] I am closing an issue 

## Description of changes 
Replace unicode right arrow with a greater-than sign. 

## Before merging

### DIFF
--- a/src/lib/components/Circos.react.js
+++ b/src/lib/components/Circos.react.js
@@ -393,13 +393,13 @@ Circos.propTypes = {
          * To display all data in the dataset use "all" as the entry for the key "name".
          * Ex: "tooltipContent": {"name": "all"}
          * 
-         * Ex: This will return (source) + ' ➤ ' + (target) + ': ' + (targetEnd)'. 
+         * Ex: This will return (source) + ' > ' + (target) + ': ' + (targetEnd)'. 
          * "tooltipContent": {
             "source": "block_id",
             "target": "position",
             "targetEnd": "value"
                                 },
-         * Ex: This will return (source)(sourceID) + ' ➤ ' + (target)(targetID) + ': ' (target)(targetEnd)'.                  
+         * Ex: This will return (source)(sourceID) + ' > ' + (target)(targetID) + ': ' (target)(targetEnd)'.                  
          * "tooltipContent": {
             "source": "source",
             "sourceID": "id",


mkcor:
Why not... But we don't want to support Python 2.7 nor encourage its use (https://pythonclock.org/)... Dash Bio is new, so I don't understand what the use case could have been. Can you please share what is was?

shammamah:
The issue came up when I was trying to add Dash Bio to the documentation, which is currently deployed on a server running Python 2.7. I believe that Dash (and its other core libraries) will continue to support Python 2.7, since it is used by some of our clients. After Python 2.7 officially "dies" in 2020, I think we can stop supporting it. (CC @alexcjohnson)

alexcjohnson:
Right - Dash itself is also new, and while I agree with you @mkcor that we don't want to *encourage* Python 2.7, it's not that simple for a lot of users. Some of it of course is inertia, but some is that our work is being used in conjunction with lots of other code, and if there is even one package a user needs that hasn't transitioned to support Python 3, they will still be using 2.7.

Aside from end-use applications for demo or single-client purposes, dropping support for 2.7 will be a company-wide decision and we are definitely not there yet. Whether we do so concurrently with the Python Clock or at some other date has yet to be determined.

mkcor:
@shammamah ok. @alexcjohnson these users can (and should) ask the maintainers of such packages for Python 3 support, and should they run into any difficulty in the process, they should reach out for help. The PSF has been working on this and providing assistance. If you know any user who currently has such dependencies (not yet ported to Python 3), I'm happy to help out.

alexcjohnson:
@mkcor at this point I'm not particularly concerned about packages folks are using within their work (though I have no idea what apache-beam is, and expect its maintainers are a bit chagrinned to be the last holdout among the top 360 packages http://py3readiness.org/). I'm more concerned about some huge internal py2 codebase at a giant company that hasn't yet convinced their management to let them spend time updating to py3. They're not going to let anyone outside their company help them with that, and they're simply not going to adopt dash if we stop supporting py2. Unless it somehow becomes a burden on us to support py2, there are still too many users stuck there for us to ignore.

Plotly is a pretty young, nimble company, and yet we're only just now porting our own internal code to py3. I guarantee many older, larger companies are not there yet.

PR #219, opened 2019-03-04
Add explicit wait time before assessing existence of elements in tests

Bachibouzouk:
## About 
Fix to prevent the ideogram tests to fail due to time loading the components

## Description of changes 
Add explicit waiting time for now so the test pass

## Before merging

### DIFF
--- a/tests/test_ideogram.py
+++ b/tests/test_ideogram.py
@@ -214,6 +214,7 @@ def test_ploidy(dash_threaded):
 
     driver = dash_threaded.driver
 
+    driver.implicitly_wait(1)
     # assert 22 chromosomes + X and Y chromosomes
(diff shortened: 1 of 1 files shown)


tests/test_ideogram.py
```diff
@@ -214,6 +214,7 @@ def assert_callback(prop_value, nclicks, input_value):
 
     driver = dash_threaded.driver
 
+    driver.implicitly_wait(1)
```
T4rk1n: Don't do that, it waits for every selenium operation after, it's not a sleep.

Bachibouzouk: Thanks for the tip. Now I use the `wait_for_elements_by_css_selector` funciton from pytest_dash instead :)

Bachibouzouk:
@shammamah : now the tests pass :)

PR #231, opened 2019-03-06
Improve demo app for VolcanoPlot

### LINKED ISSUE #165: Dash Volcano Plot: Limit range of input values
Could use a slider or a dial from Dash DAQ for example to limit the range to threshold and effect size values that make sense. If using Dash DAQ, could also use color picker for the highlight color. 

image

It also wasn't clear to me that these were outputs versus inputs:

image

Bachibouzouk:
Closes #165 

## About 
- [ ] This is a new component 
- [x] I am closing an issue 

## Description of changes 
Used dash_daq to change two Input into Slider, two Div into LEDDisplay and add a ColorPicker
## Before merging


Bachibouzouk:
The test were failing because changing the threshold value with selenium is not possible with the daq components. I created hidden Div to keep track of the values needed for the tests and rewrote one test. I erased the other one which was checking the bound values of inputs as the Slider component guarantees that the values chosen by the user is bound.

shammamah:
@Bachibouzouk Instead of hidden div, could you use `dcc.Store`? It was specifically created for this purpose (and to avoid the hidden div workaround).

Bachibouzouk:
> @Bachibouzouk Instead of hidden div, could you use `dcc.Store`? It was specifically created for this purpose (and to avoid the hidden div workaround).

@shammamah : unfortunately not, because I need to have read/write access to what is inside the hidden div with the selenium driver. I mentionned in the `className` of these divs as well as in their related callbacks that this is solely for test purpose. I can make it even clearer if you want :)

Bachibouzouk:
@shammamah : I made the modifications you suggested :)

tests/dashbio_demos/app_volcano_plot.py
```diff
@@ -113,13 +114,22 @@ def layout():
                                         "Lower effect size",
                                         className='vp-text',
                                     ),
-                                    dcc.Input(
+                                    daq.Slider(
```
shammamah: You can use a `rangeSlider` for these two as well. Just a suggestion, not super important to implement :)

tests/dashbio_demos/app_volcano_plot.py
```diff
@@ -210,20 +265,22 @@ def callbacks(app):  # pylint: disable=redefined-outer-name
...
+            Input('vp-dataset-dropdown', 'value'),
+            Input('vp-color-picker', 'value')
         ]
     )
-    def update_graph(u_lim, l_lim, genomic_line, datadset_id):
-        """Update the data set of interest upon change the dashed lines
-        value.
-        """
+    def update_graph(u_lim, l_lim, genomic_line, datadset_id, color):
+        """Update the data set of interest upon change the dashed lines value."""
```
shammamah:
```suggestion
        """Update the data set of interest upon changing the value of the dashed line."""
```

tests/dashbio_demos/app_volcano_plot.py
```diff
@@ -257,13 +343,21 @@ def update_upper_left_number(fig, l_lim):
...
+    @app.callback(
+        Output('vp-upper-right-val', 'children'),
         [Input('vp-graph', 'figure')],
         [State('vp-upper-bound', 'value')]
     )
-    def update_upper_right_number(fig, u_lim,):
+    def update_upper_right_number_val(fig, u_lim,):
         """Update the number of points in the upper right zone delimited by
-        the thresholds.
+        the thresholds for tests purpose.
```
shammamah: I'm not so sure it's a good idea to have callbacks/app features specifically so that they will work with the tests. In my opinion, it's far more important for the components to be continuously tested. Since these apps (and their code) will be on display for the public, it would be better to remove these (and, if necessary, the corresponding tests from the testing file). Otherwise, it might look confusing as these callbacks have no purpose outside of testing.

mkcor: I very much agree with @shammamah here.

Bachibouzouk: I didn't know the code of the demo would be shared publicly, my idea was that it would be good to guarantee that the app is not broken using tests so that we don't have to go every time through all the apps, testing them manually to make sure they are fine. I will simply reset the head to not include the tests then :)

shammamah: Yeah, sorry about not mentioning that! If you go to the apps on the Dash Bio portal (which is public), there is a little "View on GitHub" link at the top right. It's similar to the Dash DAQ demo apps that we created -- people should ideally be able to clone the app repo and just run the app on their own with the ability to understand what each portion of the code is doing. :)

mkcor: @Bachibouzouk I think your intention was good: If I'm not mistaken, your intention here was to program defensively, by using an ad-hoc callback in the same spirit that we include assertions, right? The idea is good, but I would see these 'tests' either as more native to the demo app, or completely separated (in a different, test-only file).

Bachibouzouk: Yes that was my intention. If the test are completely separated then if someone changes the app, the tests need to be changed again. I think we could rather use similar [functions](https://github.com/plotly/dash-cytoscape/blob/master/tests/test_interactions.py) than those @xhlulu implemented to test cytoscape to change the values of sliders. The only problem remaining is that the `daq.LEDDisplay` css doesn't contain an explicit value of the number it represents, so one would still need a hidden Div or a Store solely for test purpose.

mkcor: @Bachibouzouk Yes, but that's exactly the point of a test suite; you don't want changes that are not backward-compatible to go unnoticed... So, if they force you to update the tests, that's good, that's what we want.

tests/dashbio_demos/app_volcano_plot.py
mkcor: Really? `module 'dash_daq' has no attribute 'rangeSlider'` so we'd have to define it...

shammamah: Whoops, made a typo. It's actually a `dcc` component: https://dash.plot.ly/dash-core-components/rangeslider

mkcor: Ok, thanks!

mkcor: Unfortunately, it's not as easy as replacing `daq.Slider` with `dcc.RangeSlider`... I'm not able to integrate this [syntax](https://dash.plot.ly/dash-core-components/rangeslider) into the current design because I can't troubleshoot: Web app won't render and there's no error stream...

shammamah: Is there anything showing in the console? Usually, if the web app isn't rendering, it's due to an invalid input ID. 

My suggestion is, instead of having two `daq.Slider` components, have a single `dcc.RangeSlider` and in the callback that uses lower/upper bounds (https://github.com/plotly/dash-bio/blob/5cb2ac5cce044cc8e394217417f7b834691b7d7b/tests/dashbio_demos/app_volcano_plot.py#L264), replace the individual `daq.Slider`s with the single `RangeSlider`.

mkcor: Well, there's the expected
```
TypeError: Unexpected keyword argument `handleLabel`
Allowed arguments: allowCross, className, count, disabled, dots, id, included, loading_state, marks, max, min, pushable, step, updatemode, value, vertical
```
but after dealing with that, no more error stream... blank, void...

mkcor: > instead of having two `daq.Slider` components

I'm not sure what you mean. There are *three* Slider components, corresponding to:
sliders

mkcor: > Is there anything showing in the console?

No; to clarify, that's what I mean by "no error stream."

shammamah: I meant the two corresponding to lower/upper bounds (makes the most sense to replace with a `RangeSlider`, in my opinion). Also (sorry for not having clarified this), I'm wondering if there's anything showing up in the *browser* console. Sometimes, errors will only pop up there. 

Like I mentioned, it's not super critical to change the sliders -- if you're finding that it's taking too much time, you could leave it alone!

mkcor: Oh, I understand what you mean now!! I will leave it aside for this PR but keep it in mind for the future.

tests/dashbio_demos/app_volcano_plot.py
mkcor: So, I haven't wrapped my head around those tests; for example, it seems frustrating to convert a number into a string and then back into an integer, but I guess it's the only way to store and pass the value... @shammamah I suggest leaving this PR as is, so at least we benefit from @Bachibouzouk's improvements asap; then I guess you'll want to apply the 'standard' redesign to this demo app; and we can investigate these integration tests at some point in the future.

shammamah: @mkcor Sounds good to me!

tests/dashbio_demos/app_volcano_plot.py
```diff
@@ -106,57 +108,81 @@ def layout():
...
+                                        value=-1,
+                                        type='number',
+                                    )
                                 ],
                             ),
                             html.Div(
                                 className='vp-vertical-style',
-                                title='Changes the value of the right '
-                                      'vertical dashed line.',
+                                title='Move right-side vertical dashed line',
```
shammamah: We should figure out exactly what this dashed line represents, or at least provide a better-sounding tooltip. @mkcor what do you think?

mkcor: It's the upper bound for the effect size. Honestly, it's not my field, but I was able to understand this well from the demo app: I would it's self-contained from reading the description and looking at the plot. If you look at the app, I think it will look straightforward to you as well. Plus, you'll notice that this text is tiny, so the info wouldn't only be redundant, it also wouldn't make sense to display it in this context.

PR #232, opened 2019-03-07
Clustergram component improvements

### LINKED ISSUE #74: Clustergram component improvements
From @Bachibouzouk's comments on #70 

- [x] Use a numpy or pandas function to loop over the list when converting from list to array in `_clustergram.py` (fixed in #77)
- [x] Create a function dedicated specifically for the above 
- [ ] Ensure that row and column labels still show even if there is no clustering

shammamah:
Closes #74  

## About 
- [ ] This is a new component 
- [x] I am closing an issue 

## Description of changes 
The clustergram component actually had quite a lot of things wrong with it: 
* it was using a soon-to-be-deprecated imputer function
* it was not rearranging the rows and columns on the heatmap when there was clustering (surprised nobody caught this until now!) 
* it was losing all data about labels if there was no clustering on that dimension. (e.g., if it was clustered by "row" and the user tried to display column labels, they would simply be empty)

On top of that, the app was also changed to prevent the user from not specifying any dimensions along which to compute clustering.

## Before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [0.0.8]
 
 ### Fixed
-* Fixed issue with clustergram not reordering rows and columns after clustering.
+* Fixed issue with Clustergram not reordering rows and columns after clustering.
 
 ### Removed
 * Removed mentions of Dash events in OncoPrint component.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Changed
 * Changed property `fullChromosomeLabels` so that it can be updated using dash callbacks
 * Changed Imputer (deprecated) to SimpleImputer in Clustergram component.
+* Changed property name `impute_function` to `imputer_parameters` in Clustergram component.
 
 ### Added
 * Added ability to define custom colours in styles parser for Molecule3D.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,30 +2,34 @@
 
 ## [0.0.8]
 
+### Fixed
+* Fixed issue with clustergram not reordering rows and columns after clustering.
+
 ### Removed
 * Removed mentions of Dash events in OncoPrint component.
 
 ### Changed
 * Changed property `fullChromosomeLabels` so that it can be updated using dash callbacks
+* Changed Imputer (deprecated) to SimpleImputer for clustergram.
 
 ### Added
 * Added ability to define custom colours in styles parser for Molecule3D.
 
 ## [0.0.7] - 2019-26-02
 
-### Changed 
+### Changed
 * Changed unicode right arrow to greater-than sign in Circos for compatibility with Python 2.7.
 
 ## [0.0.6] - 2019-22-02
 
-### Added 
+### Added
 * Added requirements from files in `utils`, as well as from pure-Python components, to setup install requirements.
 * Added more descriptive prop descriptions for Dash Ideogram.
 
 ## [0.0.5] - 2019-15-02
 
-### Changed 
-* Changed filenames in `dash_bio/utils/` folder to be snake case instead of camel case. 
+### Changed
+* Changed filenames in `dash_bio/utils/` folder to be snake case instead of camel case.
 
 ## [0.0.4] - 2019-11-02
 
@@ -34,13 +38,13 @@
 
 ## [0.0.3] - 2019-06-02
 
-### Added 
-* Added variables to define strings used in `_volcano.py` graph labels.  
+### Added
+* Added variables to define strings used in `_volcano.py` graph labels.
 
 ## [0.0.2] - 2019-05-02
 
-### Fixed 
-* Fixed incompatibility issues with Dash `0.36.0`. 
+### Fixed
+* Fixed incompatibility issues with Dash `0.36.0`.
 
 ### Removed
-* Removed all mentions of `fireEvent` and anything else that used Dash events (which have been removed). 
+* Removed all mentions of `fireEvent` and anything else that used Dash events (which have been removed).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Changed
 * Changed property `fullChromosomeLabels` so that it can be updated using dash callbacks
-* Changed Imputer (deprecated) to SimpleImputer for clustergram.
+* Changed Imputer (deprecated) to SimpleImputer in Clustergram component.
 
 ### Added
 * Added ability to define custom colours in styles parser for Molecule3D.
--- a/dash_bio/component_factory/_clustergram.py
+++ b/dash_bio/component_factory/_clustergram.py
@@ -274,8 +274,8 @@ class _Clustergram():
         self._optimal_leaf_order = optimal_leaf_order
         if color_map is None:
             self._color_map = [[0.0, 'rgb(255,0,0)'],
-                              [0.5, 'rgb(0,0,0)'],
-                              [1.0, 'rgb(0,255,0)']]
+                               [0.5, 'rgb(0,0,0)'],
+                               [1.0, 'rgb(0,255,0)']]
         else:
             self._color_map = color_map
         self._color_list = color_list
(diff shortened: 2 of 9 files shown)


Bachibouzouk:
Did you rerun `npm install` before rebuilding? The failing test may be due to that...

shammamah:
@Bachibouzouk That did the trick 👍

shammamah:
(This thread says the deploy failed, but I restarted it and it works now: https://dash-bio-test-pr-232.herokuapp.com/)

CHANGELOG.md
```diff
@@ -2,30 +2,34 @@
 
 ## [0.0.8]
 
+### Fixed
+* Fixed issue with clustergram not reordering rows and columns after clustering.
```
```suggestion
* Fixed issue with Clustergram not reordering rows and columns after clustering.
```
So far, we've capitalized component names.

CHANGELOG.md
```diff
@@ -2,30 +2,34 @@
...
+
 ### Removed
 * Removed mentions of Dash events in OncoPrint component.
 
 ### Changed
 * Changed property `fullChromosomeLabels` so that it can be updated using dash callbacks
+* Changed Imputer (deprecated) to SimpleImputer for clustergram.
```
```suggestion
* Changed Imputer (deprecated) to SimpleImputer in Clustergram component.
```
If I understand correctly...

dash_bio/component_factory/_clustergram.py
```diff
@@ -61,77 +61,77 @@ def Clustergram(
...
+    :param (list/float) display_ratio: The dendrograms' heights with respect to
                                      the size of the heatmap; with one element,
                                      both the row and column dendrograms have
                                      the same ratio; with two, the row
                                      dendrogram ratio corresponds to the first
                                      element of the list (default: 0.2)
-    :param (function) imputeFunction: The function used to impute missing data.
+    :param (function) impute_function: The function used to impute missing data.
                                       It should take as input the dataset and
                                       output a dataset with imputed values.
```
mkcor: I'm not sure this description is clear enough... As is, it sounds like there's not necessarily any missing data in the original dataset, but we're imputing some... Instead, I believe that we're talking about the *representation* of missing data here. Could it be: "Function used to set the representation of missing values. It takes the whole dataset as input and outputs it with missing values converted into the desired representation."?

tests/dashbio_demos/app_clustergram.py
```diff
@@ -494,45 +494,47 @@ def update_col_threshold_max(dataset_name, contents):
...
+            'tick_font': {
                 'color': 'rgb(200,200,200)',
                 'size': 10
             },
-            'symmetricValue': dataset_name is None,
-            'logTransform': dataset_name is None,
-            'imputeFunction': {
+            'symmetric_value': dataset_name is None,
+            'log_transform': dataset_name is None,
+            'impute_function': {
```
mkcor: So, we want to change this variable name.

dash_bio/component_factory/_clustergram.py
mkcor: Ok, it's not actually a function. Will re-review after update.

dash_bio/component_factory/_clustergram.py
```diff
@@ -61,77 +61,83 @@ def Clustergram(
...
-                                      It should take as input the dataset and
-                                      output a dataset with imputed values.
-    :param (list[dict]) rowGroupMarker: Specifies which rows to annotate; each
+    :param (dict) imputer_parameters: Specifies the parameters 'missing_values'
+                                      and 'strategy' defined in the SimpleImputer
+                                      from scikit-learn 0.20.3. An additional
+                                      argument, 'axis', is used to specify the
+                                      direction along which to impute (supported
+                                      until scikit-learn 0.20). 'axis=0' indicates
+                                      that imputing should happen along columns,
```
```suggestion
                                      that imputing operation should happen along rows,
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -61,77 +61,83 @@ def Clustergram(
...
+                                      while 'axis=1' indicates that it should happen
+                                      along rows.
```
```suggestion
                                      along columns.
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -61,77 +61,83 @@ def Clustergram(
...
                                      both the row and column dendrograms have
                                      the same ratio; with two, the row
                                      dendrogram ratio corresponds to the first
                                      element of the list (default: 0.2)
-    :param (function) imputeFunction: The function used to impute missing data.
-                                      It should take as input the dataset and
-                                      output a dataset with imputed values.
-    :param (list[dict]) rowGroupMarker: Specifies which rows to annotate; each
+    :param (dict) imputer_parameters: Specifies the parameters 'missing_values'
+                                      and 'strategy' defined in the SimpleImputer
```
```suggestion
                                      and 'strategy' of function SimpleImputer
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -61,77 +61,83 @@ def Clustergram(
...
+                                      from scikit-learn 0.20.3. An additional
+                                      argument, 'axis', is used to specify the
```
```suggestion
                                      parameter, 'axis', is used to specify the
```

dash_bio/component_factory/_clustergram.py
shammamah: I'm a bit confused here -- `axis=0` indicates columns, while `axis=1` indicates rows. (Reference: https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.Imputer.html)

dash_bio/component_factory/_clustergram.py
shammamah: `SimpleImputer` here is an object, which is why I hesitate to refer to it as a function. Is this common practice?

dash_bio/component_factory/_clustergram.py
```diff
@@ -61,77 +61,83 @@ def Clustergram(
...
+                                      direction along which to impute (supported
```
```suggestion
                                      direction along which to impute (a parameter of
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -61,77 +61,83 @@ def Clustergram(
...
+                                      until scikit-learn 0.20). 'axis=0' indicates
```
```suggestion
                                      Imputer, which was deprecated in scikit-learn 0.20.0).
                                      'axis=0' indicates
```

dash_bio/component_factory/_clustergram.py
mkcor: Oh, I know where the confusion comes from! I'm used to the language found in NumPy and Pandas docs (e.g., https://docs.scipy.org/doc/numpy/glossary.html), where `axis=0` means *across* rows and `axis=1` *across* columns (also, I always scroll down to this diagram: http://swcarpentry.github.io/python-novice-inflammation/01-numpy/index.html). I find it easier to memorize, because the conventional order is to express coordinates or dimensions as [row, column] (n_row x n_column), so row is the first (index 0) and column the second (index 1) dimension... Also read this thread: https://stackoverflow.com/questions/25773245/ambiguity-in-pandas-dataframe-numpy-array-axis-definition
In this case, I'd say it's enough to reference the original parameter (all the more than it's deprecated). I would end the description after "... deprecated in scikit-learn 0.20.0)."

shammamah:
@mkcor I've addressed your feedback -- please take another look!

dash_bio/component_factory/_clustergram.py
mkcor: My bad, it's a class, as can be inferred from its name (CapWords convention). So I would suggest "parameters ... of class SimpleImputer."

mkcor: And everything is an object :)

dash_bio/component_factory/_clustergram.py
mkcor: If you don't like "imputing operation," I would suggest "imputation" instead of "imputing."
Inspiration: https://github.com/scikit-learn/scikit-learn/issues/10636

mkcor:
> @mkcor I've addressed your feedback -- please take another look!

Great! Looks like you may have missed https://github.com/plotly/dash-bio/pull/232#discussion_r264339365 ?
I have added two other suggestions following your responses. Thanks!

shammamah:
@mkcor I'm going to be fixing the docstrings as per #233 -- let's get this merged in, then I can fix all of the Python components' docstrings at once and regenerate/republish the package on TestPyPI.

mkcor:
Yes, I'm totally fine with improving docstrings iteratively!

mkcor:
I'm not sure commit 4f7a716 is fine (and that's more important than docstrings). Why `sklearn` --> `scikit-learn`? And the minimal version should be specified in `requirements.txt` rather than `setup.py`.

dash_bio/component_factory/_clustergram.py
```diff
@@ -61,77 +61,85 @@ def Clustergram(
...
+    :param (dict) imputer_parameters: Specifies the parameters 'missing_values'
+                                      and 'strategy' of the SimpleImputer class
+                                      from scikit-learn 0.20.3. An additional
+                                      parameter, 'axis', is used to specify the
+                                      direction along which to impute (a parameter of
+                                      Imputer, which was deprecated in scikit-learn 0.20.0).
+                                      'axis=0' indicates that imputing should happen along columns,
+                                      while 'axis=1' indicates that it should happen
+                                      along rows (see:
+                                      http://lijiancheng0614.github.io/scikit-learn/modules/
```
mkcor: Wait, why this personal webpage and not the official scikit-learn docs?

shammamah:
>Why sklearn --> scikit-learn?

This is actually the library we should have been using the whole time (reference: https://pypi.org/project/sklearn/)

>the minimal version should be specified in requirements.txt rather than setup.py.
The requirements file isn't included in the package (AFAIK), whereas `setup.py` enforces that the correct versions are on the machine at the time that `dash_bio` is installed.

dash_bio/component_factory/_clustergram.py
shammamah: Oops -- copy/pasted the  wrong link.

mkcor:
> > Why sklearn --> scikit-learn?
> 
> This is actually the library we should have been using the whole time (reference: https://pypi.org/project/sklearn/)

Good to know!

> > the minimal version should be specified in requirements.txt rather than setup.py.
> 
> The requirements file isn't included in the package (AFAIK), whereas `setup.py` enforces that the correct versions are on the machine at the time that `dash_bio` is installed.

Ok, so we're treating dash_bio as an app, not as a library we're distributing (which is the original purpose of PyPI).

shammamah:
>Ok, so we're treating dash_bio as an app, not as a library we're distributing (which is the original purpose of PyPI).

Not really following here. Dash does the same thing, even though it is a library (https://github.com/plotly/dash/blob/e255ba0ff3ce5e3080b596ee04f9e5c902a07953/setup.py#L19).

shammamah:
(Let's continue this discussion over on #135!)

PR #245, opened 2019-03-18
Sequenceviewer redesign

shammamah:
## About 
Sequence viewer app redesign. 

## Description of changes 
* Move all controls to `dcc.Tabs` 
* Move sequence info underneath sequence viewer component and group the two in a single parent div 
* Fix heights of both sequence info container and sequence viewer component container
* Change accent colour 

## Before merging

### DIFF
--- a/tests/dashbio_demos/app_sequence_viewer.py
+++ b/tests/dashbio_demos/app_sequence_viewer.py
@@ -357,90 +357,91 @@ def layout():
                         ]),
 
                         html.Div(id='seq-view-sel-slider-container', children=[
-                                html.Div(className='seq-view-controls-name',
-                                         children="Selection region:"),
-                                dcc.RadioItems(
-                                    id='sel-slider-or-input',
-                                    options=[
-                                        {'label': 'slider', 'value': 'slider'},
-                                        {'label': 'input', 'value': 'input'}
-                                    ],
-                                    value='slider'
-                                ),
-                                dcc.RangeSlider(
-                                    id='sel-slider',
-                                    min=0,
-                                    max=0,
-                                    step=1,
-                                    value=[0, 0]
-                                ),
-                                # optional numeric input for longer sequences
-                                html.Div(
-                                    id='sel-region-inputs',
-                                    children=[
-                                        "From: ",
-                                        dcc.Input(
-                                            id='sel-region-low',
-                                            type='number',
-                                            min=0,
-                                            max=0,
-                                            placeholder="low"
-                                        ),
-                                        "To: ",
-                                        dcc.Input(
-                                            id='sel-region-high',
-                                            type='number',
-                                            min=0,
-                                            max=0,
-                                            placeholder="high"
-                                        ),
-                                    ],
-                                    style={'display': 'none'}
-                                ),
+                            html.Div(
+                                className='seq-view-controls-name',
+                                children="Selection region:"
+                            ),
+                            dcc.RadioItems(
+                                id='sel-slider-or-input',
+                                options=[
+                                    {'label': 'slider', 'value': 'slider'},
+                                    {'label': 'input', 'value': 'input'}
+                                ],
+                                value='slider'
+                            ),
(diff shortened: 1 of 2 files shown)


tests/dashbio_demos/app_sequence_viewer.py
```diff
@@ -69,197 +69,298 @@ def layout():
...
 
-                html.Div(
-                    id='sequence-viewer-container',
+        html.Div(id='seq-view-control-tabs', children=[
+            dcc.Tabs(id='seq-view-tabs', children=[
+                dcc.Tab(
+                    label='About',
+                    value='what-is',
+                    children=html.Div(className='seq-view-tab', children=[
+                        html.H4('What is Sequence Viewer?'),
```
mkcor: I don't think this renders very well, plus what follows doesn't answer the question in a way that would be syntactical (it doesn't start with "Sequence Viewer is ..." but with an invitation to "Display ..."). So I would remove this level-4 header. See my comment in the general conversation as well.

mkcor:
Why pink? I guess why not!

I took a screenshot of how the demo app renders for me:
responsive

Since it's not responsive, I think we want to ensure our fixed format looks good: The line break between 'Sequence' and 'Viewer' doesn't look good (in "What is Sequence Viewer?"). As I commented in the review, I think the whole "What is Sequence Viewer?" isn't welcome anyway. "Dash Sequence Viewer" at the top-left corner is prominent enough.

shammamah:
>Why pink? 

Thought it might be fun to have a pink app! The pink is also the exact shade of the number inside the little "badge". (I had thought it would be red, but apparently not.) If it's too garish, I can easily change it. 

> The line break between 'Sequence' and 'Viewer' doesn't look good

Hm, weird that it looks that way on yours - on my computer, it's all on one line (and, as you mentioned, all of the sizes/widths are fixed). I'll make the text size smaller! 

> I think the whole "What is Sequence Viewer?" isn't welcome anyway.

Personally I'd prefer to keep it in, just because it follows the pattern of the other apps (e.g., https://dash-bio.plotly.host/dash-alignment-viewer/). This might also be a problem that will be solved by making the text size smaller.

mkcor:
Pink is great, don't change it!

I didn't see such a level-4 header in Speck, and I thought it had the new standard layout, so that's what I used as a reference point.

Well, what follows "What is Alignment Viewer?" is "The Alignment Viewer (MSA) component is used ..." which is perfectly correct. On the contrary, what follows "What is Sequence Viewer?" is not consistent.

shammamah:
>I find that the zone for the sequence is too small compared to the description zone (cf the screenshot of my view).

I'm not quite sure what this means -- do you mean that more lines should be visible in the sequence viewer component? 

>when one uses the input method for the start and stop of the sequence selection, the selected sequence starts from start_idx +1.

The sequence is indexed like an array/list -- if you start the selection from 0, it will include the first character. Do you think this will be confusing for people? (If so, I can change it)

>For the choice of the color of the background and text, could we make use of a dash_daq colorpicker or it is too big? (It would be nice to have to possibility to have the color picker appear only if the user press a button and be replaced by the button upon color choice. The button could have the color set as the background with no text or with RGB color code in white)

Yeah, I tried this out and found that they were too big. Good idea re: selectively displaying the colorpicker. I think that can be handled in a future PR, though -- let's get this redesign merged/deployed ASAP! 

>Suggestion for further improvements :

These are all awesome ideas! Please open an issue -- that way, I can keep track of them.

shammamah:
@mkcor Good call -- longer description of the app has been added in f192343.

mkcor:
> @mkcor Good call -- longer description of the app has been added in [f192343](https://github.com/plotly/dash-bio/commit/f192343c4cc9365a682b258874269986de18e611).

Wonderful! Note that, although 'data set' is not wrong, the trend is to spell it 'dataset' (and I think the last whitespace in 'acid composition in the box under the component. ' wasn't intended).

PR #273, opened 2019-03-23
Exclude react-* packages from Renovate updates

mkcor:
Addresses https://github.com/plotly/dash-bio/pull/272#issuecomment-475899788.

## About 
This is about dependency management.

## Description of changes
Reference: https://renovatebot.com/docs/configuration-options/#excludepackagepatterns

## Before merging

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
     "node":
         docker:
-            - image: circleci/node:8.11.3
+            - image: circleci/node:11.10.1
 
         steps:
             - checkout
--- a/renovate.json
+++ b/renovate.json
@@ -21,6 +21,9 @@
   "lockFileMaintenance": {
     "enabled": false
   },
+    "packageRules": [{
+      "excludePackagePatterns": ["react-*"]
+  }],
   "ignoreDeps": [
     "ideogram",
     "react"


mkcor:
Wondering if upgrading `node` will also upgrade `npm`... If not, I'll add `npm i npm@latest -g` to the CI config.

shammamah:
@mkcor It might be a good idea to add it to the CI config, just in case!

mkcor:
> Actually, I'm a bit concerned about this line:
> 
> [dash-bio/package.json](https://github.com/plotly/dash-bio/blob/d7bb8ae048496cbeebc377bff10794f01ec0ca23/package.json#L28)
> 
> Line 28 in [d7bb8ae](/plotly/dash-bio/commit/d7bb8ae048496cbeebc377bff10794f01ec0ca23)
>  "react-oncoprint": "^1.2.2", 
> 
> If we are excluding all `react-*` packages, won't that also prevent some of the other dependencies (the ones we need for our components), too?

Good catch. Note that excluding *automated updates* for certain packages doesn't deprive us of these dependencies, and doesn't prevent us from updating them manually at any moment (as we have done so far). For `Ideogram`, we have locked the dependency... so we may want to control the updates of other component-specific dependencies. If you don't think so, I'm happy to write out `react-dom`; note that we can always add or remove packages to be updated *automatically* (again, we are not removing dependencies, just selecting which ones could be updated automatically).

mkcor:
> @mkcor It might be a good idea to add it to the CI config, just in case!

Great, will do in a separate PR!

mkcor:
> Actually, I'm a bit concerned about this line:

So, basically, there's nothing to be concerned with, since this PR doesn't remove any functionality wrt what we've been doing so far. Note that, so far, we haven't merged any PR submitted by the Renovate Bot which contained grouped updates!

shammamah:
OK, all of that sounds good to me. :)

mkcor:
> [upgrading npm in CI config]
> 
> Great, will do in a separate PR!

Won't do in the end. Reference is @Marc-Andre-Rivet's sharing the following: "When installing `node` you will get a specific version of `npm` — but you can update `npm` independently afterwards. Pulling the latest version of `npm` in your toolchain probably won’t cause any problem but, like any other dependency, it may break or have changes in behavior — your CI docker pulls a specific version of `node`, hence `npm`, I would stick to that and instead update the referenced docker image version if/when necessary." Thank you!!

Issue #260, opened 2019-03-20
Ideogram test for number of chromosomes fails intermittently.


mkcor:
We've had this CI fail with Python 3.7 over and over again, so there must be a bug:
![intermittent_fail](https://user-images.githubusercontent.com/2227806/54697539-08d0ef00-4b2e-11e9-9ca2-75c65e212391.png)

We want this number of chromosomes to be deterministic.

mkcor:
Sorry, this intermittent fail also happens with Python 3.6:
![fail_py36](https://user-images.githubusercontent.com/2227806/54708481-bdc1d680-4b43-11e9-8420-bbdcc649ea84.png)

mkcor:
My understanding of `test_chromosomes_wrong_input` (https://github.com/plotly/dash-bio/blob/master/tests/test_ideogram.py#L297) is that a wrong chromosome name fails silently... Since 'D' isn't valid, it is simply ignored and only '1' and '3' are retained.

mkcor:
What I've noticed is that, in the screenshots above and CI failures we keep seeing, `num_chromosoms` actually retained its former value instead of updating as expected. Therefore, I'll try increasing the value of parameter `timeout` in the `wait_for` function.

mkcor:
This other example illustrates the above comment https://github.com/plotly/dash-bio/issues/260#issuecomment-476273274, where the second region wasn't found within 10 seconds (default timeout value):
![region_count](https://user-images.githubusercontent.com/2227806/55184110-a614ce80-5191-11e9-8119-860e59583fc4.png)

mkcor:
Closing for now, because we haven't seen these failures lately (and PR #280, although no silver bullet, may improve things). Will reopen if necessary.

PR #280, opened 2019-03-28
Update ideogram tests.

mkcor:
Adresses #260. 

## About
Certain integration tests count (expected) elements by having Selenium find them (CSS selector). Sometimes, the underlying function times out before updated elements could be found and, hence, counted. 

## Description of changes
Delving into the Ideogram tests, I have standardized the finding / counting of elements. I have pushed frequently to have 'statistics' on CI test failures, and lately they haven't failed... So I'm thinking of not changing 
```
-    chromosomes = wait_for_elements_by_css_selector(driver, '.chromosome')
+    chromosomes = wait_for_elements_by_css_selector(driver, '.chromosome', timeout=20.0)
```
(at least, not yet).
## Before merging

### DIFF
--- a/src/lib/components/Ideogram.react.js
+++ b/src/lib/components/Ideogram.react.js
@@ -8,7 +8,7 @@ import {omit} from 'ramda';
  * datasets for organisms such as human, mouse, and any other
  * eukaryote. The Ideogram component can be used to compare
  * homologous features between chromosomes, and depict
- * haploid, diploid, aneuploidy genomes. It can also display
+ * haploid, diploid, aneuploid genomes. It can also display
  * annotations on genomic data using histograms and overlays.
  *
  * Reference: https://eweitz.github.io/ideogram/
--- a/tests/test_ideogram.py
+++ b/tests/test_ideogram.py
@@ -20,7 +20,7 @@ from .test_common_features import (
 # This version is locked in `package-lock.json`. If that changed and some of the tests failed,
 # it might be due to changes in https://github.com/eweitz/ideogram. For example, test_orientation
 # fails with version v1.5.1 because "chromosome-set-container" was renamed to "chromosome-set"
-# (which results in a timeout of the _wait_for function).
+# (which results in the wait_for function timing out without finding the element).
 
 # define app name once
 APP_NAME = os.path.basename(__file__).replace('test_', '').replace('.py', '').replace('_', '-')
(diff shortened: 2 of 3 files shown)


mkcor:
PS: Letting the `plotly` dependency be `>=3.5.0` (in practice, I'm now running the latest version `3.7.1`) doesn't seem to change anything. I have explored the demo apps locally. There's still this error in the console when accessing Dash Alignment Viewer:
```
TypeError: the JSON object must be str, bytes or bytearray, not NoneType
```

src/lib/components/Ideogram.react.js
```diff
@@ -8,7 +8,7 @@ import {omit} from 'ramda';
  * datasets for organisms such as human, mouse, and any other
  * eukaryote. The Ideogram component can be used to compare
  * homologous features between chromosomes, and depict
- * haploid, diploid, aneuploidy genomes. It can also display
+ * haploid, diploid, aneuploid genomes. It can also display
```
shammamah: Words like these make me remember why I switched my major from biochem 🤣

tests/test_ideogram.py
```diff
@@ -227,16 +226,16 @@ def assert_callback(prop_value, nclicks, input_value):
     driver = dash_threaded.driver
 
     # assert 22 chromosomes + X and Y chromosomes
-    num_chromosoms = len(wait_for_elements_by_css_selector(driver, '.chromosome'))
-    assert num_chromosoms == 24
+    chromosomes = wait_for_elements_by_css_selector(driver, '.chromosome')
```
shammamah: Is there a particular reason that you changed these? (I'm just wondering if it somehow speeds things up, or ensures that we have a result before trying to test the assertion)

tests/test_ideogram.py
```diff
@@ -629,7 +636,7 @@ def assert_callback(prop_value, nclicks, input_value):
 
     # assert the presence of homology region
     regions = wait_for_elements_by_css_selector(driver, '.syntenicRegion')
-    assert len(regions) != 0
+    assert len(regions) > 0
```
shammamah: Why was this changed? I guess it makes sense, since the length of a list cannot be negative, but I'm not sure this change affects performance or functionality. (Correct me if I'm wrong!)

tests/test_ideogram.py
Bachibouzouk: I think it is to standardize the tests?

mkcor: For both reasons: standardize and get elements before taking their length (for possible debugging too).

tests/test_ideogram.py
mkcor: It shouldn't affect performance or functionality, but if ever a length was negative, I would want to catch it!! Well, I'm for writing the meaningful version directly (we expect this length to be positive, so let's write it, it's easier on the reader too).

PR #296, opened 2019-04-01
Fix babel error and update ideogram documentation. 

shammamah:
Building the bundle after the dependencies updated from @renovate-bot (https://github.com/plotly/dash-bio/pull/284) was resulting in the following error: 
```
ERROR in ./src/lib/index.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: Cannot find module '@babel/core'
 babel-loader@8 requires Babel 7.x (the package '@babel/core'). If you'd like to use Babel 6.x ('babel-core'), you should install 'babel-loader@7'.
```

We can also try to work with the changes Renovate Bot has suggested, instead of fixing the `babel-loader` package to an outdated version (@mkcor -- would love your thoughts on this, since you've been working the most with the Renovate Bot.) 

## Description of changes 
* Fix `babel-loader` version to `7.0.0` so that babel-core can be used.
* Generate new version of Ideogram with the updated docstrings from #ideogram

## Before merging

### DIFF
--- a/dash_bio/Ideogram.py
+++ b/dash_bio/Ideogram.py
@@ -9,7 +9,7 @@ The Ideogram component is used to draw and animate genome-wide
 datasets for organisms such as human, mouse, and any other
 eukaryote. The Ideogram component can be used to compare
 homologous features between chromosomes, and depict
-haploid, diploid, aneuploidy genomes. It can also display
+haploid, diploid, aneuploid genomes. It can also display
 annotations on genomic data using histograms and overlays.
 
 Reference: https://eweitz.github.io/ideogram/
--- a/dash_bio/metadata.json
+++ b/dash_bio/metadata.json
@@ -1183,7 +1183,7 @@
     }
   },
   "src/lib/components/Ideogram.react.js": {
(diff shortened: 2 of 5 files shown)


mkcor:
Good catch on updating the Ideogram Python class and metadata! Proof that I didn't rebuild the `dash_bio` package...

mkcor:
@shammamah I would much prefer upgrading Babel (https://babeljs.io/docs/en/v7-migration) rather than downgrading `babel-loader`. Apparently, Babel 7 has been around for over half-a-year (https://babeljs.io/blog/2018/08/27/7.0.0), so I think it would be about time to migrate... Now, if this migration turns out to be very demanding, we can pass in the very short term (and I can approve this PR), but please try first and let me know how much effort you estimate it would require.

shammamah:
Fixed the babel presets and the dev dependencies as per the instructions here: https://github.com/babel/babel/issues/6808#issuecomment-343792786

mkcor:
> Fixed the babel presets and the dev dependencies as per the instructions here: [babel/babel#6808 (comment)](https://github.com/babel/babel/issues/6808#issuecomment-343792786)

Yay! Great job, @shammamah :+1:

PR #241, opened 2019-03-14
Speck setprops fix

shammamah:
## About 
- [ ] This is a new component 
- [x] I am adding a feature to an existing component, or improving an existing feature
- [ ] I am closing an issue 

## Description of changes 
Remove all "unprotected" calls to  `setProps`, which is only defined as a function if there is a callback in the Dash app that uses the component as input. Also ensure that `view` has a full set of parameters at any given time (before, it would be restricted to the dictionary supplied by the user and there would be errors related to not being able to compute the rendering because of missing properties). 

## Before merging
- [x] I have gone through the [code review checklist](https://github.com/plotly/dash-component-boilerplate/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/review_checklist.md)
- [x] I have completed all of the [steps to take before merging](https://github.com/plotly/dash-bio/blob/master/.github/before_merging.md)
- [x] I know how to [deploy to DDS](https://github.com/plotly/dash-bio/blob/master/.github/deploying_to_dds.md)

PR #297, opened 2019-04-01
Dash renderer update

shammamah:
(note that I am not publishing a new version of the package, since the problem with Speck's `view` prop from #241 still needs to be fixed)

## About 
Updates to `dash` and `dash-renderer` ensure that `setProps` does not have to be defined by creating a callback that takes the component as input. This fixes part of the issue referenced in #241. 

## Description of changes 
* Change the Dash version in the requirements for the demo apps and in the setup install requirements for Dash Bio. 
* Change the demo apps to have default tabs, as the newest version of `dash-core-components` does not display the contents of the first tab by default.
* Make the IDs for the tabs in each of the demo apps more uniform. 
* Remove unnecessary callback in Speck application.

## Before merging

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,6 @@ dash-bio==0.0.8rc13
 dash-core-components==0.43.1
 dash-daq==0.1.4
 dash-html-components==0.13.5
-dash-table==3.4.0
 gunicorn
 jsonschema
 matplotlib
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,12 +5,11 @@
 biopython
 colour==0.1.5
 cython>=0.19
-dash==0.37.0
+dash>=0.40.0
 dash-bio==0.0.8rc13
 dash-core-components==0.43.1
 dash-daq==0.1.4
 dash-html-components==0.13.5
-dash-renderer==0.18.0
 dash-table==3.4.0
 gunicorn
 jsonschema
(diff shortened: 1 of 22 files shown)


setup.py
```diff
@@ -18,7 +18,7 @@
     install_requires=[
         'biopython',
         'colour',
-        'dash>=0.37.0',
+        'dash>=0.40.0',
         'dash-html-components',
```
mkcor: Can we remove this dependency, then? Just like you did in `requirements.txt`.

requirements.txt
```diff
@@ -5,13 +5,9 @@
 biopython
 colour==0.1.5
```
mkcor: Doesn't have to be part of this PR, but do you know if we could loosen this one into `>=` as well?

shammamah: We should be able to. `0.1.5` is actually the latest version, so I don't anticipate there being any huge breaking changes in the near future. It's only used in `dash_bio/utils/circos_parser.py` regardless. (Plus, in the `setup.py` install requirements, a version isn't specified so it will default to the latest version anyway.)

shammamah: Let's leave that to a different PR, since you had mentioned needing to change the other dependencies to `>=` as well. (We can do all of them at once!)

setup.py
shammamah: @mkcor Done in 213d238 :)

requirements.txt
mkcor: Actually, I mentioned all the dependencies in the *Dash ecosystem*, so I think we've covered that :)

PR #301, opened 2019-04-02
Clustergram improvements

### LINKED ISSUE #288: APP QA 2: Clustergram
- [ ] Colorbar needs a label (is it cm?)
- [x] "Petal length(cm)" (missing a space)
- [x] Dropdowns here are smushed
image
Need to change this:
image
Looks much better:
image

- [ ] Would be cool if had control to adjust levels of hierarchy in tree
- [x] Can't figure out add/remove group markers UI. What does group markers mean? 
- [ ] Can we coordinate zoom/drag on subplots?
image
- [ ] Not sure if this is possible with plotly.js but would be cool if heatmaps had a max zoom-out, so you couldn't do things like this:
image
- [x] There's vertical gap here:
image
- [ ] Put a hyperlink here:
image
- [x] This dropdown is messed:
image
- [x] x button shows up on top of text - make dropdown much longer:
image
- [ ] Plot needs a title
- [x] Default datasets should be oncology or genetics focused... Might be able to find data from an open journal article
- [x] Dropdown is touching horizontal rule here:
image

shammamah:
Addresses part of #288 

## Description of changes 
* Add ability in the Clustergram component to produce a dictionary that can translate the `clickData` from `dcc.Graph` into the cluster that was clicked.
* Add options to return the above dictionary, as well as an option to use precomputed traces, into the Clustergram component. Both of these are false by default. 
* Return *only* the figure data by default. This avoids the need to unpack values if one only wishes to display the clustergram.
* Add date to CHANGELOG for version `0.0.8`. 
* Refine and make use of the option to use precomputed traces in the demo application.

## Before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [0.0.9]
 
+### Fixed
+* Fixed issue with Clustergram rows and columns reordering incorrectly
+  on the heatmap when precomputed traces are used to generate the
+  figure.
+
 ### Changed
 * Changed Clustergram to only return a figure by default, so that
   values no longer need to be unpacked.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
   values no longer need to be unpacked.
 
 ### Added
-* Added to Clustergram the ability to generate a "curves dictionary"
+* Added to Clustergram the ability to generate a "curve dictionary"
 to translate curve number (available in hoverData/clickData) to the
 row or column cluster it represents on the graph.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 # Changelog
 
-## [0.0.8]
+## [0.0.9]
+
+### Changed
+* Changed Clustergram to only return a figure by default, so that
+  values no longer need to be unpacked.
+
+### Added
+* Added to Clustergram the ability to generate a "curves dictionary"
+to translate curve number (available in hoverData/clickData) to the
+row or column cluster it represents on the graph.
+
+## [0.0.8] - 2019-01-04
 
 ### Fixed
 * Fixed issue with Clustergram not reordering rows and columns after clustering.
--- a/dash_bio/component_factory/_clustergram.py
+++ b/dash_bio/component_factory/_clustergram.py
@@ -850,8 +850,8 @@ Keyword arguments:
     def _compute_clustered_data(
             self
     ):
-        """Gets the traces that need to be plotted for the row and column
-        dendrograms, and updates the ordering of the 2D data array,
+        """Get the traces that need to be plotted for the row and column
+        dendrograms, and update the ordering of the 2D data array,
         row labels, and column labels to match the reordered
         dendrogram leaves.
 
--- a/dash_bio/component_factory/_clustergram.py
+++ b/dash_bio/component_factory/_clustergram.py
@@ -14,6 +14,8 @@ from plotly import tools
 # pylint: disable=assignment-from-no-return, no-self-use
 def Clustergram(
         data=None,
+        generate_curves_dict=False,
+        use_precomputed_traces=False,
         computed_traces=None,
         row_labels=None,
         column_labels=None,
@@ -48,6 +50,15 @@ def Clustergram(
 Keyword arguments:
 
 - data (ndarray; required): Matrix of observations as array of arrays
+- generate_curves_dict (bool; optional): Whether or not to return a
+    dictionary containing information about the cluster number
+    associated with each curve number in the graph. (May be useful
+    if one wishes to capture the cluster number that is clicked.)
+    (Default: False)
+- use_precomputed_traces (bool; optional): Whether or not to return
+    the precomputed dendrogram traces. (May be useful if one wishes
+    to add, e.g., group markers to the figure without recalculating
+    the clustering in the entire figure.) (Default: False)
 - computed_traces (dict; optional): The dendrogram traces from another
    Clustergram component.
 - row_labels (list; optional): List of row category labels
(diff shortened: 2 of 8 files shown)


dash_bio/component_factory/_clustergram.py
```diff
@@ -799,14 +847,25 @@ def _get_clusters(
 
         return (Zcol, Zrow)
 
-    def _dendrogram_traces(
+    def _compute_clustered_data(
             self
     ):
         """Gets the traces that need to be plotted for the row and column
-        dendrograms.
+        dendrograms, and updates the ordering of the 2D data array,
```
```suggestion
        dendrograms, and update the ordering of the 2D data array,
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -799,14 +847,25 @@ def _get_clusters(
 
         return (Zcol, Zrow)
 
-    def _dendrogram_traces(
+    def _compute_clustered_data(
             self
     ):
         """Gets the traces that need to be plotted for the row and column
```
```suggestion
        """Get the traces that need to be plotted for the row and column
```

tests/dashbio_demos/app_clustergram.py
```diff
@@ -130,7 +135,28 @@ def layout():
                     label='About',
                     value='what-is',
                     children=html.Div(className='clustergram-tab', children=[
-                        html.P('Gene expression data and heatmap with dendrogram.')
+                        html.H3('What is Clustergram?'),
+                        html.P('Clustergram is a combination of a heatmap and '
+                               'dendrograms that allows you to display '
+                               'hierarchical clustering data. '
+                               'Clusters on the dendrograms are highlighted in '
+                               'different colors if they comprise data points '
```
mkcor:
```suggestion
                               'similar colors if they comprise data points '
```
I'm not sure, but that's what I would expect logically, if/since they "share some minimal level of correlation."

tests/dashbio_demos/app_clustergram.py
```diff
@@ -130,7 +135,28 @@ def layout():
...
+                               'that share some minimal level of correlation.'),
+                        html.P('In the "Data" tab, you can select a preloaded '
+                               'dataset to display or, optionally, upload one '
```
mkcor:
```suggestion
                               'dataset to display or, alternatively, upload one '
```

tests/dashbio_demos/app_clustergram.py
```diff
@@ -130,7 +135,28 @@ def layout():
...
+                               'of your own. A sample dataset is also available '
+                               'for download in the tab.'),
+                        html.P('In the "Graph" tab,  you can choose the '
```
mkcor:
```suggestion
                        html.P('In the "Graph" tab, you can choose the '
```

tests/dashbio_demos/app_clustergram.py
```diff
@@ -285,62 +306,40 @@ def layout():
...
+                                    n_clicks_timestamp=0
                                 ),
-                                dcc.Input(
-                                    id='color',
-                                    placeholder='color',
-                                    type='text',
-                                    value=''
+                                html.Br(),
+                                html.Div(className='clustergram-option-desc', children=[
+                                    'Annotate your heatmap by labelling clusters; '
```
mkcor:
```suggestion
                                    'Annotate your heatmap by labeling clusters; '
```
US spelling, unfortunately ;)

tests/dashbio_demos/app_clustergram.py
```diff
@@ -285,62 +306,40 @@ def layout():
...
+                                    'choose a color for the annotation, as well as '
+                                    'text for the annotation, below. Then, click '
```
mkcor: Hmm, I would move 'below' to "choose below a color ..."

tests/dashbio_demos/app_clustergram.py
```diff
@@ -637,7 +659,7 @@ def display_clustergram(
                 }
             )
         if fig_opts['cluster'] is None:
-            return html.Div(
+            wrapper_content = html.Div(
                 'No clustering dimension has been selected to display. Please \
```
mkcor: Isn't it "no dimension has been selected to perform clustering along?"

tests/dashbio_demos/app_clustergram.py
shammamah: You're right -- good catch! The colors are actually the exact same (that's what you see in the app -- the different-colored "trees" correspond to different clusters. I'll reword this part to reflect that :)

tests/dashbio_demos/app_clustergram.py
shammamah: Yep, that sounds clearer.

CHANGELOG.md
```diff
@@ -1,6 +1,22 @@
...
+* Fixed issue with Clustergram rows and columns reordering incorrectly
+  on the heatmap when precomputed traces are used to generate the
+  figure.
+
+### Changed
+* Changed Clustergram to only return a figure by default, so that
+  values no longer need to be unpacked.
+
+### Added
+* Added to Clustergram the ability to generate a "curves dictionary"
```
mkcor: I think it should be referred to as a "curve dictionary" (just like a real adjective, the "noun as adjective" is invariable; reference I could find very quickly: https://www.englishclub.com/grammar/nouns-adjective.htm).

tests/dashbio_demos/app_clustergram.py
shammamah: I actually find the way that I have it clearer -- is it conventional to structure the sentence in the way you've suggested?

mkcor: I think the more conventional way would be to write "below, choose a color" (look it up, but I've never seen 'below' isolated at the very end of a sentence, with that separating comma: ", below.").

mkcor:
@shammamah the addition of file `tests/dashbio_demos/#app_speck.py#` is suspicious... might you be using `git commit -a` by any chance?

shammamah:
@mkcor Something similar -- was using `git add tests/dashbio_demos`. I'll avoid doing this in the future!

tests/dashbio_demos/app_clustergram.py
shammamah: Changed in cd7b7b0!

PR #304, opened 2019-04-04
Build actual dash_bio node.js package in CI build.

### LINKED ISSUE #300: Package building isn't part of our CI.
Whenever we change JS code, we rebuild the package by running
```bash
npm run build:all
```
which results in `dash_bio/bundle.js`. If this breaks, we have a broken package. Therefore, it is important to make this command part of CI (it's currently [missing](https://github.com/plotly/dash-bio/blob/fc254890310b03502952c6f6c4a430f77538b9bf/.circleci/config.yml#L80) in our CI config).

Not only do we want to double check builds that were successful locally (maybe our dev environments aren't fully up-to-date and/or we don't run both Python 3.6 and Python 3.7), but we also want PRs from the Renovate Bot (dependency updates, such #299 and #284) to go through the full pipeline.

mkcor:
Closes #300. 

## About
We wanted to ensure that `dash_bio/bundle.js` builds successfully. This is now done, with the changes submitted here.

## Description of changes
I've added command `npm run build:js` to the "node" job, which runs in a node.js environment.

Ultimately, we may want to address the following warning:
size_limit_warning

## Before merging

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,18 @@ jobs:
                 command: npm run build:js
 
             - run:
+                name: "Create virtualenv"
+                command: |
+                    python -m venv venv
+
+            - run:
+                name: "Install Dash"
+                command: |
+                    . venv/bin/activate
+                    pip install --upgrade pip
+                    pip install dash==0.40.0
+
+            - run:
                 name: "Build Python package"
                 command: npm run build:py
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,10 @@ jobs:
                 command: ./node_modules/.bin/eslint src
                 when: always
 
+            - run:
+                name: Build package
+                command: npm run build:all
+
 
     "python-3.6": &test-template
         docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,15 +43,12 @@ jobs:
                     python -m venv venv
 
             - run:
-                name: "Install Dash"
+                name: "Build Python package"
                 command: |
                     . venv/bin/activate
                     pip install --upgrade pip
                     pip install dash==0.40.0
-
-            - run:
-                name: "Build Python package"
-                command: npm run build:py
+                    npm run build:py
 
 
     "python-3.6": &test-template
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,13 +38,9 @@ jobs:
                 command: npm run build:js
 
             - run:
-                name: "Create virtualenv"
-                command: |
-                    python -m venv venv
-
-            - run:
                 name: "Build Python package"
                 command: |
+                    python -m venv venv
                     . venv/bin/activate
                     pip install --upgrade pip
                     pip install dash==0.40.0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,8 +26,8 @@ jobs:
                 when: always
 
             - run:
-                name: Build package
-                command: npm run build:all
+                name: Build node.js package
+                command: npm run build:js
 
 
     "python-3.6": &test-template
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ jobs:
 
             - run:
                 name: "Build Node.js package"
-                command: npm run build:js
+                command: npm run build:js-dev
 
             - run:
                 name: "Build Python package"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,10 +3,18 @@ version: 2
 jobs:
     "node":
         docker:
-            - image: circleci/node:11.10.1
+            - image: circleci/python:3.7-stretch-node
 
         steps:
             - checkout
+            - run:
+                name: "Update Node.js and npm"
+                command: |
+                  curl -sSL "https://nodejs.org/dist/v11.10.1/node-v11.10.1-linux-x64.tar.xz" | sudo tar --strip-components=2 -xJ -C /usr/local/bin/ node-v11.10.1-linux-x64/bin/node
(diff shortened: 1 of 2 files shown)


mkcor:
The Python classes are built via `npm run build:py` -- I wonder how/where we could run this command...

shammamah:
@VeraZab Is there anything that we can do about the size of the bundle? We're already using the minified code.

shammamah:
@mkcor `npm run build:all` runs `build:js` and `build:py` -- so that's the only one that we need to run :)

.circleci/config.yml
```diff
@@ -25,6 +25,10 @@ jobs:
                 command: ./node_modules/.bin/eslint src
                 when: always
 
+            - run:
+                name: Build node.js package
+                command: npm run build:js
```
shammamah:
```suggestion
                command: npm run build:all
```

mkcor: This doesn't run (look at the commit history of this PR), which is expected, since the node environment doesn't have any python deps...

mkcor:
> @VeraZab Is there anything that we can do about the size of the bundle? We're already using the minified code.

Well, first we should study https://webpack.js.org/guides/code-splitting/ in depth--has any one of us already done so? I haven't.

shammamah:
Right, that makes sense. Is there a way to install the Python requirements before running the node job? Here
https://github.com/plotly/dash-bio/blob/d5dadfedf4752c4994d6fab02235e38f8f1e2a0c/.circleci/config.yml#L95
I'm seeing that you can define jobs and then run them in a specific order: https://circleci.com/docs/2.0/workflows/

So we can just create a new job in our CircleCI config that installs `dash` (that should be the only package used to build the Python components), and run that before we run `npm run build:all`.

mkcor:
> @mkcor `npm run build:all` runs `build:js` and `build:py` -- so that's the only one that we need to run :)

As is, we have separate node and python envs, so we can't run `npm run build:js & npm run build:py` at once.

> Misread the script name.

When reviewing a PR, it's useful to look at the Commits tab, not just the diff (Files tab).

mkcor:
> @shammamah So we can just create a new job in our CircleCI config that installs dash

No, each *job* has its own docker image, as you can see. So if you want to use a dependency that was just installed (`dash` in our case), these two commands have to run within the same *job*.

I'm happy to install `dash` before running `npm run build:py` but I think we should keep the two commands (`npm run build:js & npm run build:py`) so the sequence is explicit.

shammamah:
Sure, I think that makes the most sense.

mkcor:
@shammamah wait, I think it's really not straightforward to run Python code in this Node environment... We are now using a Docker image that's readily available:
https://github.com/plotly/dash-bio/blob/c90e6b139689ac4b75f0896233d2664d65398f41/.circleci/config.yml#L6
We would need a different (custom, Python-equipped) one, should we run the following Python code:
```py
import dash; dash.development.component_loader.generate_classes('dash_bio', 'dash_bio/metadata.json')
```
as done by `npm run build:py`:
https://github.com/plotly/dash-bio/blob/c90e6b139689ac4b75f0896233d2664d65398f41/dash_bio/package.json#L13
I'm hesitating between two options:

* creating a dedicated Docker image suitable for running `npm run build:py`;
* breaking down the various commands contained in `npm run build:py` (doable by resorting to `store_artifacts`).

Gets me thinking a lot about the node/python coupling...

shammamah:
>creating a dedicated Docker image suitable for running npm run build:py;

I'd choose this, since it represents a well-defined "step" in our build process.

VeraZab:
@mkcor @shammamah this looks like a reasonable change to me.
I feel like you guys know better at this point, since I haven't kept updated about the latest evolvements of the repo.

mkcor:
I went for a Node.js variant Docker image, including our own specific version of Node.js / NPM as exemplified here: https://circleci.com/docs/2.0/circleci-images/#best-practices
This way, we can compile the components into Python right after building the Node.js package (9c9a4bd) and after installing `dash` (7d4a1a2).

mkcor:
I thought caching the `dash` dependency wasn't worthwhile, hence 18010a798d2fcd5c.

mkcor:
Oh, and I probably need to cherry-pick this commit https://github.com/plotly/dash-bio/pull/312/commits/1b324503cae514641bc059f57eb40266002470b3 :)

.circleci/config.yml
```diff
@@ -3,10 +3,18 @@ version: 2
...
         steps:
             - checkout
+            - run:
+                name: "Update Node.js and npm"
+                command: |
+                  curl -sSL "https://nodejs.org/dist/v11.10.1/node-v11.10.1-linux-x64.tar.xz" | sudo tar --strip-components=2 -xJ -C /usr/local/bin/ node-v11.10.1-linux-x64/bin/node
+                  curl https://www.npmjs.com/install.sh | sudo bash
+            - run:
+                name: Check current version of node
+                command: node -v
```
shammamah: What does this command do? Is it just for logging purposes?

.circleci/config.yml
```diff
@@ -25,6 +33,19 @@ jobs:
                 command: ./node_modules/.bin/eslint src
                 when: always
 
+            - run:
+                name: "Build Node.js package"
+                command: npm run build:js
```
shammamah:
```suggestion
                command: npm run build:js-dev
```

This will catch the errors, but it is significantly faster (since it doesn't have to minify the code).

.circleci/config.yml
mkcor: Yes, it only checks for the current version of node, as the `name` says :)
It may be unnecessary, but I copied-pasted and adapted from https://circleci.com/docs/2.0/circleci-images/#best-practices

shammamah: No harm in leaving it in! It will probably end up being useful in the future. :)

PR #312, opened 2019-04-10
Speck system fix

shammamah:
## Description of changes 
Previously, the Speck component was trying to calculate the radius for an atom even if no atoms were present in the system. This led to an exception and a failure to render. 

Now, the function `loadStructure` simply returns if there are no atoms supplied in the system and consequently avoids this error.

I also fixed the CHANGELOG to have nicer formatting and added the fix from #311 to the notes for the 0.0.9 release.

## Before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@
   figure.
 * Fixed issue with Speck not rendering unless it is attached to a callback.
 * Prevent Speck from trying to calculate a system with no atom.
-  which led to an error.
 
 ### Changed
 * Changed Clustergram to only return a figure by default, so that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,59 +6,72 @@
 * Fixed issue with Clustergram rows and columns reordering incorrectly
   on the heatmap when precomputed traces are used to generate the
   figure.
+* Fixed issue with Speck not rendering unless it is attached to a callback.
+* Fixed issue with Speck trying to calculate a system with no atoms,
+  which led to an error.
 
 ### Changed
 * Changed Clustergram to only return a figure by default, so that
   values no longer need to be unpacked.
 
 ### Added
-* Added to Clustergram the ability to generate a "curve dictionary"
-to translate curve number (available in hoverData/clickData) to the
-row or column cluster it represents on the graph.
+* Added to Clustergram the ability to generate a "curve dictionary" to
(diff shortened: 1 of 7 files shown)


shammamah:
The previous deployments were failing due to #313. I've added a temporary fix by locking the Dash version to `0.40.0`, which uses the last version of `dash` that accepts `React.PropTypes`.

CHANGELOG.md
```diff
@@ -6,64 +6,78 @@
...
+  using dash callbacks.
+* Changed Imputer (deprecated) to SimpleImputer in Clustergram
+  component.
+* Changed property name `impute_function` to `imputer_parameters` in
+  Clustergram component.
 * Changed install requirement to Dash version 0.40.0 or greater.
 
 ### Added
-* Added ability to define custom colours in styles parser for Molecule3D.
+* Added ability to define custom colours in styles parser for
```
```suggestion
* Added ability to define custom colors in style parser for
```

CHANGELOG.md
```diff
@@ -6,64 +6,78 @@
 * Fixed issue with Clustergram rows and columns reordering incorrectly
   on the heatmap when precomputed traces are used to generate the
   figure.
+* Fixed issue with Speck not rendering unless it is attached to a callback.
+* Fixed issue with Speck trying to calculate a system with no atoms,
```
```suggestion
* Prevent Speck from trying to calculate a system with no atom.
```

CHANGELOG.md
```diff
@@ -6,64 +6,78 @@
...
+  which led to an error.
```
mkcor: I suggest removing this line.

shammamah: @mkcor This is an issue that was fixed, though -- do you have an idea of how it could be worded better? (Or does this belong in another section?)

mkcor: My rewording suggestion corresponds to making the whole sentence fit in one line, which is "Prevent Speck from trying to calculate a system with no atom." It says it all. We fixed an issue (when there was an issue, there was an error, but it's trivial so there's no need to specify it; it's not surprising that trying to calculate an empty system would lead to an error; it's not interesting to write out "there used to be an error" and the source code is clear and commented...).

"Prevent Speck from trying to calculate a system with no atom" makes it clear that *previously* Speck was trying to calculate such a system and clearly that was an issue!!

shammamah: Thanks for the clarification :) Fixed in 8a1f343!

PR #315, opened 2019-04-11
Volcanoplot redesign

shammamah:
## Before merging
- [x] I have gone through the [code review checklist](https://github.com/plotly/dash-component-boilerplate/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/review_checklist.md)
- [x] I have completed all of the [steps to take before merging](https://github.com/plotly/dash-bio/blob/master/.github/before_merging.md)
- [x] I know how to [deploy to DDS](https://github.com/plotly/dash-bio/blob/master/.github/deploying_to_dds.md)

PR #322, opened 2019-04-12
Fix broken downloads

### LINKED ISSUE #320: Sample data download locations are incorrect in portal apps
One of them is in `assets/`, and one of them is in `tests/dashbio_demos`. The one in `assets` is more recent, so the other one should be removed.

shammamah:
Closes #320 

## About 
- [ ] This is a new component 
- [x] I am closing an issue 

## Description of changes 
* Keep sample data files in `/assets/sample_data/` to distinguish them from the preloaded datasets that are in `/tests/dashbio_demos/sample_data/`. 
* Use `os.path.join` instead of a string to refer to the location of sample data. 

## Before merging


mkcor:
> Keep sample data files in `./assets/sample_data/` to distinguish them from the preloaded datasets that are in `./tests/dashbio_demos/sample_data/`

Can you please clarify? Since both subfolders are named `sample_data/`, I would expect both to contain some kind of *sample data*. Now, maybe one kind of 'sample data' is available for download via the UI, while the other isn't... Is that the right distinction?

shammamah:
@mkcor Yes, that's correct. We could rename the downloadable sample data folder to `downloadable_sample_data`, maybe?

mkcor:
> `downloadable_sample_data`

Maybe `downloadable_data` would suffice... depending on where it lives. That would be the one under `./assets/`, right? But, yes, that would be a great move!

shammamah:
Yeah, that's correct. The download links should all say "Download sample [data type] data", so that's why I'm inclined to keep "sample" in the name.

mkcor:
Ok, I agree. Also because `assets/` is at the root of the repo, so let's be more explicit / explanatory.

mkcor:
Do you prefer merging #315 or this one first?

shammamah:
Let's merge #315 in. I'd rather update all of the sample data locations at once in a separate PR!

mkcor:
Ok. But, actually, why are these guys https://github.com/plotly/dash-bio/pull/315/commits/8537df9382671c008d4db81c8c6cb54a25b5e9bb#diff-8a48fd800a995577d334b7602a8b3e97already in #315?

shammamah:
Honestly not quite sure -- I was probably testing the downloads on the wrong branch. I'll revert those commits -- good catch!

Issue #289, opened 2019-03-31
App QA 2: Manhattan plot


jackparmer:
- [x] Notice that these threshold lines are at different levels when the app loads. They shift downward
![image](https://user-images.githubusercontent.com/1865834/55284805-e1f39380-534b-11e9-8628-4d70072e6d4c.png)
I think the graph initial load values for these lines is different from the defaults set in the slider callbacks.. then the slider callbacks get called on Dash app load so shift downwards... Just a guess on what is going on
- [x] Slider -> graph is still too slow to update... Taking 5 seconds. Can we use Dash clientside here? https://github.com/plotly/dash/pull/672 Can we use a much smaller dataset/graphs so that is doesn't take 5 seconds to redraw on? Would also be cool if we could put in a loading spinner on update
- [x] Would be nice if these chromosomes were spaced out a bit:
![image](https://user-images.githubusercontent.com/1865834/55284873-33505280-534d-11e9-9465-9e44a0c2b26b.png)
- [ ] Autoscale is a bit weird. Can we limit the zoom-out upper bound? Probably don't also need these 2 axes lines:
![image](https://user-images.githubusercontent.com/1865834/55284876-609d0080-534d-11e9-91ab-f2dd1b5d98ef.png)

PR #326, opened 2019-04-22
Trim down data before plotting.

shammamah:
Addresses part of #289 

## Description of changes 
Trim down the dataset to use only the first 100 rows of the data for each chromosome in order to speed up load/response times for the graph. 

## Before merging

### DIFF
--- a/tests/dashbio_demos/app_manhattan_plot.py
+++ b/tests/dashbio_demos/app_manhattan_plot.py
@@ -94,7 +94,7 @@ def layout():
                             value=2,
                             marks={
                                 i + 1: '{}'.format(i + 1)
-                                for i in range(9)
+                                for i in range(4)
                             },
                             step=0.05
                         ),
@@ -114,7 +114,7 @@ def layout():
                             value=3,
                             marks={
                                 i + 1: '{}'.format(i + 1)
-                                for i in range(9)
+                                for i in range(4)
                             },
                             step=0.05
                         )
--- a/tests/dashbio_demos/app_manhattan_plot.py
+++ b/tests/dashbio_demos/app_manhattan_plot.py
@@ -18,7 +18,13 @@ elif 'DASH_PATH_ROUTING' in os.environ:
 DATAPATH = os.path.join(".", "tests", "dashbio_demos", "sample_data", "manhattan_")
 
 # Load the data
-DATASET = pd.read_csv("{}data.csv".format(DATAPATH))
+DT = pd.read_csv("{}data.csv".format(DATAPATH))
+
+# Trim down the data
+datasets = [DT.loc[DT['CHR'] == i+1] for i in range(23)]
+datasets = [dataset.iloc[:100] for dataset in datasets]
+
+DATASET = pd.concat(datasets).reset_index(drop=True)
 
 # Feed the data to a function which creates a Manhattan Plot figure
 fig = dash_bio.ManhattanPlot(DATASET)
@@ -83,9 +89,9 @@ def layout():
                             id='mhp-slider-genome',
                             vertical=False,
                             updatemode='mouseup',
-                            max=9,
+                            max=4,
                             min=1,
-                            value=7,
+                            value=2,
                             marks={
                                 i + 1: '{}'.format(i + 1)
                                 for i in range(9)
@@ -103,9 +109,9 @@ def layout():
                             id='mhp-slider-indic',
                             vertical=False,
                             updatemode='mouseup',
-                            max=9,
+                            max=4,
                             min=1,
-                            value=6,
+                            value=3,
                             marks={
                                 i + 1: '{}'.format(i + 1)
                                 for i in range(9)
--- a/tests/dashbio_demos/app_manhattan_plot.py
+++ b/tests/dashbio_demos/app_manhattan_plot.py
@@ -1,4 +1,5 @@
 import os
+
 import pandas as pd
 import dash_html_components as html
 import dash_core_components as dcc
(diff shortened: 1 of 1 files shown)


shammamah:
@mkcor Are you referring to the preprocessing of the data set? It doesn't have anything to do with the component -- the users of this app will never be modifying that dataset. I trimmed it down in the app instead of altering it directly in case we decide to implement Dash client-side or something equivalent that will allow for faster processing for the whole dataset.

mkcor:
I know. I'm not referring to the component either. I'm talking programming very generally. You want to be more defensive. For example, please let this `23` come out of somewhere... If we know these datasets are human genomes, great, but then take the max of the given variable (column) and ensure it equals 23.

mkcor:
Also ensure that variable 'CHR' is present in the dataframe before using it. Et cetera, et cetera.

tests/dashbio_demos/app_manhattan_plot.py
```diff
@@ -18,7 +18,13 @@
 DATAPATH = os.path.join(".", "tests", "dashbio_demos", "sample_data", "manhattan_")
 
 # Load the data
-DATASET = pd.read_csv("{}data.csv".format(DATAPATH))
+DT = pd.read_csv("{}data.csv".format(DATAPATH))
+
+# Trim down the data
+datasets = [DT.loc[DT['CHR'] == i+1] for i in range(23)]
```
mkcor: Out of curiosity, wondering if `DT.where(DT.CHR == i + 1)` would be equivalent...

shammamah: I don't think so -- `DataFrame.where` *replaces* the value in the dataframe with `NaN` (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.where.html). Here, I'm trying to split the data by chromosome, then trimming down the data for each chromosome separately, then putting them back together to result in a smaller dataset.

shammamah:
@mkcor Your comments have been addressed in 47e505f and d14aa67!

tests/dashbio_demos/app_manhattan_plot.py
mkcor: Oh, right, it's a mask. Thanks for double checking!

tests/dashbio_demos/app_manhattan_plot.py
```diff
@@ -83,12 +91,12 @@ def layout():
                             id='mhp-slider-genome',
                             vertical=False,
                             updatemode='mouseup',
-                            max=9,
+                            max=4,
```
mkcor: Does `4` happen to be `DATASET.shape[1]` or something?

tests/dashbio_demos/app_manhattan_plot.py
```diff
@@ -18,7 +18,15 @@
...
-DATASET = pd.read_csv("{}data.csv".format(DATAPATH))
+DT = pd.read_csv("{}data.csv".format(DATAPATH))
+
+# Trim down the data
+if 'CHR' in DT:
+    datasets = [DT.loc[DT['CHR'] == i+1] for i in range(DT['CHR'].max())]
+    datasets = [dataset.iloc[:50] for dataset in datasets]
+    DATASET = pd.concat(datasets).reset_index(drop=True)
+else:
+    DATASET = None
```
mkcor: I don't think that's very satisfying... We want to return some kind of warning message at the very least. I would suggest replacing this if/else with either a try/except or `assert 'CHR' in DT`, in order to fire an error when `'CHR`' isn't available in the dataset (because then the demo app isn't usable anyway).

tests/dashbio_demos/app_manhattan_plot.py
shammamah: No, it's based on the y-values of the data (9 was too high -- there were no points at that value).

tests/dashbio_demos/app_manhattan_plot.py
shammamah: We've done the same thing with other apps, since we know which datasets they will be using. (e.g., https://github.com/plotly/dash-bio/blob/51a88da4c10decf1421b63fc14bf4be21767ffd1/tests/dashbio_demos/app_clustergram.py#L67) I don't think it's actually necessary to do this sort of thing, since the warning message will never pop up (because, again, these apps were made with specific datasets in mind, and we can guarantee that that specific dataset has a column named `CHR`). I can do something similar to what I did with clustergram above, and define these values in a dictionary -- but either way, they'll end up hardcoded.

tests/dashbio_demos/app_manhattan_plot.py
mkcor: I had to dig up the component and the data to figure out this y-value must be `'ZSCORE'`. So, it wouldn't hurt to take the max and then ceiling of that variable (plus some padding if desirable)... so the {reader, reviewer, reuser} of this piece of code has some idea of what is going on.

mkcor:
It's not strictly about hard-coding. Of course, demo apps are specific and they are not meant to be very flexible or portable (by definition). Nonetheless, when you read code (and that includes your future self), you want to read some kind of story. Usually, the structure should be reminiscent of other stories and the content should be explicit (in this case, we make it clear that the data is about the 23 human chromosome pairs).

As I rearranged your code in the spirit of data exploration bc57b89, I realized you were reinventing the wheel in terms of the split-apply-combine approach (https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html), so I've implemented it b112277. Please note you need to update `pandas` to version `0.24.2` to run it. Is it fine by you?

shammamah:
@mkcor Nice! I had no idea that you could do that with Pandas. That is totally OK by me.

PR #327, opened 2019-04-22
Improve Clustergram datasets.

shammamah:
Addresses part of #288 

## Before merging


dash_bio/utils/gene_expression_reader.py
```diff
@@ -1,57 +1,151 @@
...
 import pandas as pd
 import tempfile
 
 
 # pylint: disable=unnecessary-lambda
 
-def parse_tsv(
+def read_soft_file(
         contents='',
         filepath='',
```
mkcor: This looks weird to me. It looks like we use this function by specifying either `contents` or `filepath`, is that correct? Then, I understand why the other argument doesn't matter and comes with a meaningless default value... But I don't find it very clean. Let's take sides and favour either `contents` or `filepath`.

dash_bio/utils/gene_expression_reader.py
```diff
@@ -1,57 +1,151 @@
...
+        return_filtered_data=False
+):
     desc = {}
-    info = pd.read_csv(filepath, sep='^', nrows=header_rows-1, header=None)[0]
-    for i in info:
-        tmp = i.strip('#').split(':', 1)
-        if len(tmp) < 2:
-            desc['Source'] = tmp[0]
+
+    if len(contents) > 0:
```
mkcor: We may also want to make sure it's a character string.

dash_bio/utils/gene_expression_reader.py
```diff
@@ -1,57 +1,151 @@
...
+        all_rows,
+        all_cols,
+        rows,
+        columns
+    )
 
+
+def read_tsv_file(
+        contents='',
+        filepath='',
```
mkcor: Same comment as above.

dash_bio/utils/gene_expression_reader.py
shammamah: So if we favour contents, for example, would we simply ignore `contents` if the filepath is not `None`?

dash_bio/utils/gene_expression_reader.py
```diff
@@ -1,57 +1,151 @@
...
+        all_rows,
+        all_cols,
+        rows,
+        columns
+    )
+
+
+def _get_selected_data(
+        dataframe,
+        all_rows=None,
```
mkcor:
```suggestion
        all_rows=[],
```

dash_bio/utils/gene_expression_reader.py
```diff
@@ -1,57 +1,151 @@
...
+        all_cols=None,
```
mkcor:
```suggestion
        all_cols=[],
```

dash_bio/utils/gene_expression_reader.py
```diff
@@ -1,57 +1,151 @@
...
+        rows=None,
```
mkcor:
```suggestion
        rows=[],
```

dash_bio/utils/gene_expression_reader.py
```diff
@@ -1,57 +1,151 @@
...
+        columns=None
```
mkcor:
```suggestion
        columns=[]
```

dash_bio/utils/gene_expression_reader.py
```diff
@@ -1,57 +1,151 @@
...
+):
+    if all_rows is None:
+        all_rows = []
+    if all_cols is None:
+        all_cols = []
+    if rows is None:
+        rows = []
+    if columns is None:
+        columns = []
```
mkcor: And remove all the above if's.

dash_bio/utils/gene_expression_reader.py
shammamah: I think this is actually an antipattern: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

dash_bio/utils/gene_expression_reader.py
```diff
@@ -1,57 +1,151 @@
...
+
+    selected_rows = []
```
mkcor:
```suggestion
    selected_rows = list(set(all_rows).intersection(rows))
```
and discard the following loops... no?

shammamah: Good catch!

shammamah: Fixed in 59260c4

dash_bio/utils/gene_expression_reader.py
shammamah: Let's do this in a subsequent PR. There are actually a few parsers that use this pattern -- e.g., https://github.com/plotly/dash-bio/blob/96d43e90c4598b42282ee27b28e827c8e0f6ec7e/dash_bio/utils/protein_reader.py#L29 
I'd prefer to get those all done in one go, if that's OK!

dash_bio/utils/gene_expression_reader.py
shammamah: Let's do this in a subsequent PR. There are actually a few parsers that use this pattern -- e.g., https://github.com/plotly/dash-bio/blob/96d43e90c4598b42282ee27b28e827c8e0f6ec7e/dash_bio/utils/protein_reader.py#L29 
I'd prefer to get those all done in one go, if that's OK!

mkcor: If we favour contents, `contents=''` would become `contents` (has to be specified) and `filename=''` could be fine (since it's meaningless, it might as well be `filename=None`). Alternatively, if standard, conventional names exist (but I guess we would mean basename rather than filename), we can use them as default values. I'm not familiar with the format, but I guess along the lines of `contents='data', filepath='data.soft'`.

Okay to change it in a dedicated PR.

shammamah: Great -- please open an issue for that in the repo!

mkcor: Done #330 -- resolving then.

dash_bio/utils/gene_expression_reader.py
mkcor: Oh, yes, thanks! I think these docs are great but really confusing for calling their example function `append()` (which it does while it "isn't meant to").

dash_bio/utils/gene_expression_reader.py
mkcor: Actually, we'll take care of it as we address #330.

PR #331, opened 2019-04-25
Specify pandas version in requirements.

shammamah:
## Description of changes 
Deployment to DDS fails because the version of `pandas` is not specified. If there is already a version of `pandas` installed in the container, it will not update (because `pip` will recognize the package as already having been installed). We didn't run into this problem with the demo apps because the heroku containers are ephemeral and consequently will always have the latest version of `pandas`.

## Before merging

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ gunicorn
 jsonschema
 matplotlib
 numpy
-pandas
+pandas>=0.24.2
 parmed
 plotly>=3.5.0
 requests


mkcor:
Ok, but I think it raises the question of how containers are gardened on DDS... We would want to mimic (catch) it on Heroku and/or reflect it in our CI where, on the contrary, we reinstall Python dependencies every time, since we fire up a container every time (https://github.com/plotly/dash-bio/blob/e9b42bb6d24d81e0f9300956a418310fd5b70824/.circleci/config.yml#L99).

shammamah:
@cldougl I remember this being an issue before, too -- is there a way to ensure that a new container is created for every deployment?

mkcor:
Alternatively, we could add a human / manual step to the deployment process. But why not specifying versions for all Python dependencies, after all? That's typically what the `requirements.txt` file is for (back to the `setup.py` / `requirements.txt` confusion discussion).

shammamah:
Yep, I think that is the safest thing to do here. I'll open an issue for it!

PR #334, opened 2019-04-26
Take snapshot of Python environment

### LINKED ISSUE #332: Python dependency versions are not specified in requirements.txt
Alternatively, we could add a human / manual step to the deployment process. But why not specifying versions for all Python dependencies, after all? That's typically what the `requirements.txt` file is for (back to the `setup.py` / `requirements.txt` confusion discussion).

_Originally posted by @mkcor in https://github.com/plotly/dash-bio/pull/331#issuecomment-486691953_

mkcor:
Closes #332 

## About 
With this change, we shouldn't run into any more surprises when deployment with DDS. That being said, I don't understand why the current version of Dash would be `>=0.40.0` (https://github.com/plotly/dash-bio/blame/master/requirements.txt didn't enlighten me): I 'reverted' `dash>=0.40.0` to `dash==0.40.0` because, if I remember correctly, we need to upgrade React to v16 before we can upgrade our Dash dependency.

## Description of changes 
I have used package versions from my local dev environment (output of `pip freeze`).

## Before merging

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,19 +2,19 @@
 
 # Switch into a virtual environment
 # pip install -r requirements.txt
-biopython
+biopython==1.73
 colour==0.1.5
 cython>=0.19
-dash>=0.40.0
+dash==0.40.0
 dash-bio==0.0.9-rc10
 dash-daq==0.1.4
-gunicorn
-jsonschema
-matplotlib
-numpy
+gunicorn==19.9.0
+jsonschema==2.6.0
+matplotlib==3.0.2
+numpy==1.15.4
 pandas>=0.24.2
-parmed
+parmed==3.1.0
 plotly>=3.5.0
-requests
+requests==2.21.0
 scipy>=1.1.0
-scikit-learn
+scikit-learn==0.20.2


shammamah:
@mkcor The library still works with Dash 0.42.0 -- we've been building the package and running the apps with the `>=` dependency without any issues. We should change the rest of them to `>=` as well, to avoid potential conflicts in the future.

mkcor:
> @mkcor The library still works with Dash 0.42.0 -- we've been building the package and running the apps with the `>=` dependency without any issues. We should change the rest of them to `>=` as well, to avoid potential conflicts in the future.

Are you sure that Dash 0.42.0 was running then? I'm currently running the test suite locally to check. If Dash 0.40.0 was installed, it's logical and expected that `dash>=0.40.0` didn't create any problem, since Dash 0.40.0 is compatible with the requirement `dash>=0.40.0`. But it doesn't ensure that Dash 0.42.0 ever ran in the containers that we use for deployment... (as we recently discussed over at #331)

mkcor:
@shammamah Update: so far *all the tests* are failing (with Dash 0.42.0). We haven't upgraded our React yet...

shammamah:
Oh, I see what you mean here: 
https://github.com/plotly/dash-bio/blob/790f488df05a0cdbe7df2bdb1340d9d09fc4e74a/.circleci/config.yml#L40

I had thought that this would also be a `dash >= 0.40.0`, and consequently the package build would succeed with `0.42.0`. Sorry about that!

PR #340, opened 2019-05-01
Update homepage on PyPI.

### LINKED ISSUE #339: Dash Bio homepage is wrong on PyPI
The link currently leads to a repository titled `dash_bio`, which is nonexistent.

shammamah:
Closes #339 

## Before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
-## [0.0.9]
+## [0.0.10] - 2019-04-30
+
+### Fixed
+* Fixed homepage link in package metadata.
+
+## [0.0.9] - 2019-04-29
 
 ### Fixed
 * Fixed issue with Clustergram rows and columns reordering incorrectly
--- a/dash_bio/package.json
+++ b/dash_bio/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dash-bio",
-  "version": "0.0.9",
+  "version": "0.0.10",
   "description": "Dash components for bioinformatics",
   "main": "build/index.js",
   "scripts": {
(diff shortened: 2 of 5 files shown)


shammamah:
@mkcor OK -- added an entry. I'm wondering if we need to add CHANGELOG entries for every single patch (that seems a little excessive, especially when it's something as minor as this that doesn't affect how the package behaves at all).

shammamah:
For example, the CHANGELOG for Dash (https://github.com/plotly/dash/blob/master/CHANGELOG.md) is only updated for minor versions.

mkcor:
Ok. I think you're misinterpreting what I said. It's the other way around.

I'm only asking for a CHANGELOG entry because you bumped up the version number. Personally, I would have made this tiny change without changing versions... and all this (including WIP docs) would have been smoother and faster.

But if you go from 0.0.9 to 0.0.10 (which I don't think was necessary, I actually think it's a bit odd), somehow you have to justify it... right? I still don't understand why you bumped version number to 0.0.10, but now that you did, I'm trying to make it work somehow.

shammamah:
Actually, you *do* have to bump the version number to edit the metadata. See discussion here: https://github.com/pypa/warehouse/issues/2170#issuecomment-352902581

shammamah:
More info here: https://mail.python.org/pipermail/distutils-sig/2017-December/031826.html

mkcor:
Looking at the CHANGELOG for Dash, it doesn't seem related to anything major/minor/patch/rc. It looks like it's updated whenever there's a version updated, whether minor or patch or anything.

shammamah:
Ah, you're correct! Didn't scroll down far enough :)

mkcor:
That's super interesting, thanks for sharing. But here I'm not really arguing about whether to increment versions or not. Rather, I'm arguing that *if you do so*, then you want to document it.

Do you mind rerunning CI to get those tests to pass?

Issue #337, opened 2019-04-29
Can't get dash-bio bundle


mdelmans:
Getting the following when trying to render SequenceViewer component from master branch:

GET https://unpkg.com/dash_bio/dash_bio/bundle.js net::ERR_ABORTED 404

shammamah:
Can you please post the code that you were using?

mdelmans:
``` python
from flask import Flask

from dash import Dash
from dash_bio import SequenceViewer
import dash_html_components as html

server = Flask(__name__) 

app = Dash(server = server, name = "SequenceViewer", share = True, url = '/seqview/')

app.layout = html.Div([
	SequenceViewer(id='seqview', sequenceMaxHeight='100px')
])

if __name__ == "__main__":
	server.run(debug=True)

```

shammamah:
Try adding in this line: 

```python
app.scripts.config.serve_locally = True
```

This is because the Dash Bio JS distribution is not yet hosted on NPM. 

Side question -- is there a particular reason that you're using Flask? You can instead use Dash directly: 

```python
from dash import Dash
from dash_bio import SequenceViewer
import dash_html_components as html

app = Dash(__name__)

server = app.server

app.layout = html.Div([
    SequenceViewer(id='seqview', sequenceMaxHeight='100px')
    ])

if __name__ == "__main__":
    app.run_server(debug=True)
```

mdelmans:
Thanks, that worked. I'm using Flask because I host many apps on the same server.

PR #347, opened 2019-05-08
Publish to NPM

shammamah:
Addresses part of #337. cc @Marc-Andre-Rivet 

## About
Deployment for dash_bio doesn't work unless the line `app.scripts.config.serve_locally = True` is included in the app file. This is because there is currently no NPM distribution for the package. 

Instead of using the `.npmignore` file, I used the `files` property in `package.json` to specify the only files that should be included (only `dash_bio/bundle.js`). However, due to the original `build:py` script copying `package.json` from the root directory into `dash_bio`, the `files` property was specified within that directory too and since there is no `dash_bio/dash_bio/bundle.js` file, the tarball for the NPM package did not include `dash_bio/bundle.js` at all. 

After having published, the code in https://github.com/plotly/dash-bio/issues/337#issuecomment-490113307 no longer results in an error. 

## Description of changes
* Only captured relevant version info (required for `setup.py` and/or `dash_bio/__init__.py`) into a file `dash_bio/package_info.json` instead of copying the entire `package.json` file into the `dash_bio` directory. 
* Used "files" property in `package.json` instead of `.npmignore`. 
* Modified setup and init files accordingly to work with this new structure. 

## Before merging


Marc-Andre-Rivet:
@rpkyle This change is similar to `dash-table` in terms of impact for DashR. Would you prefer this to match the table's naming or for us to move away from the `package` nomenclature entirely as we are aiming to a more permanent / globally applicable solution.

@shammamah This looks fine to me.

alexcjohnson:
yes please, call it `package-info.json`, not `package_info.json`, for the time being while we sort out how to standardize/fix this behavior. That matches `dash-table` as well as the naming convention of `package-lock.json`

PR #353, opened 2019-05-15
Fixes for exceptions encountered while running the demo apps

mj3cheun:
## About
- [ ] This is a new component
- [X] I am adding a feature to an existing component, or improving an existing feature
- [ ] I am closing an issue

## Description of changes
Fixes some exceptions encountered while trying to run demo apps.

## Before merging


tests/dashbio_demos/app_clustergram.py
```diff
@@ -652,12 +652,14 @@ def add_marker(
...
+                current_group_markers[
+                    '{}_group_marker'.format(cluster_dimension)
+                ].append(marker)
             except ValueError:
                 pass
-
-            current_group_markers[
-                '{}_group_marker'.format(cluster_dimension)
-            ].append(marker)
+            except UnboundLocalError:
```
shammamah: Can you clarify exactly when this error would be raised? I don't think we're using any global variables within this function.

mj3cheun: yes basically in the if statement on line 644 `cluster_number` and `cluster_dimension` are defined if true and undefined if not true. The `UnboundLocalError` exception is required to ensure errors don't get thrown in either case.

Issue #354, opened 2019-05-16
Dash Speck resets to default view state with Dash Loading States


shammamah:
**Describe the bug**
Updating the view parameters from the presets results in no corresponding change in the component.

**To Reproduce**
- Change to the "View" tab
- Select "Use presets"
- Change atom style or rendering style

**Expected behavior**
Atom style or rendering style should change. 

**Python version: 3.7.2**

**Python environment (all installed packages in your current environment):**
```
appnope==0.1.0
astroid==2.2.0
atomicwrites==1.3.0
attrs==18.2.0
backcall==0.1.0
biopython==1.73
bleach==3.1.0
certifi==2018.11.29
chardet==3.0.4
chromedriver-binary==2.45.0
Click==7.0
colour==0.1.5
cycler==0.10.0
Cython==0.29.6
dash==0.40.0
dash-bio==0.0.11
dash-bio-utils==0.0.0rc1
dash-core-components==0.45.0
dash-daq==0.1.4
dash-html-components==0.15.0
dash-renderer==0.21.0
dash-table==3.6.0
decorator==4.3.2
docutils==0.14
flake8==3.6.0
Flask==1.0.2
Flask-Compress==1.4.0
GEOparse==1.1.0
gunicorn==19.9.0
idna==2.8
ipdb==0.11
ipython==7.3.0
ipython-genutils==0.2.0
isort==4.3.9
itsdangerous==1.1.0
jedi==0.13.3
Jinja2==2.10
jsonschema==2.6.0
jupyter-core==4.4.0
kiwisolver==1.0.1
lark-parser==0.6.6
lazy-object-proxy==1.3.1
MarkupSafe==1.1.1
matplotlib==3.0.2
mccabe==0.6.1
more-itertools==6.0.0
nbformat==4.4.0
numpy==1.15.4
pandas==0.24.2
ParmEd==3.1.0
parso==0.3.4
percy==2.0.1
pexpect==4.6.0
pickleshare==0.7.5
pkginfo==1.5.0.1
plotly==3.5.0
pluggy==0.9.0
prompt-toolkit==2.0.9
ptyprocess==0.6.0
py==1.8.0
pycodestyle==2.4.0
pyflakes==2.0.0
Pygments==2.3.1
pylint==2.3.0
pypandoc==1.4
pyparsing==2.3.1
pyrsistent==0.14.11
pytest==4.3.0
pytest-dash==2.0.0
python-dateutil==2.8.0
pytz==2018.9
[...]

PR #355, opened 2019-05-16
Speck React refactoring

mj3cheun:
## About
- [ ] This is a new component
- [X] I am adding a feature to an existing component, or improving an existing feature
- [ ] I am closing an issue

## Description of changes
Refactors react wrapper for speck

## Before merging


src/lib/components/Speck.react.js
```diff
@@ -8,70 +8,11 @@ import {
     speckPresetViews,
 } from 'speck';
 
-export default class Speck extends Component {
```
shammamah: We use this structure in our other libraries (e.g., https://github.com/plotly/dash-core-components/blob/master/src/components/Input.react.js). I think it's OK to keep it for now if it's not affecting the performance or functionality in any way. If this is a convention that is widely used, however, it would be worthwhile to look into implementing it everywhere (but that's a discussion with a larger scope than just this component). 

cc @alexcjohnson

shammamah: In light of the above comment, could you please change this back to the way it was, so that the diffs are clearer?

shammamah:
Another small thing -- if this closes the issue #354, please note that in the initial PR comment! (There should be some text there that you can fill in.) GitHub can auto-close the issue once this one is merged in.

src/lib/components/Speck.react.js
mj3cheun: I only chose to do it this way because the linter wants me to keep `React` in scope and keeping that in mind using `React.Component` results in the least amount of code lol. I can change it back.

shammamah:
Hey! Did you publish the new package on PyPI? I don't see a `0.0.12rc1` release. Please make sure that you publish it, and then change the `requirements.txt` file at the root of the repository to reflect the newest version. This way, the demo app that gets auto-built and deployed will give us a chance to play with the behaviour and make sure it all still works. 

Please also fix the conflicts in this branch!

dash_bio/package.json
```diff
@@ -0,0 +1,65 @@
+{
```
shammamah: Please remove this file!

src/lib/components/Speck.react.js
```diff
@@ -243,7 +218,7 @@ Speck.propTypes = {
         atomScale: PropTypes.number,
         relativeAtomScale: PropTypes.number,
         bondScale: PropTypes.number,
-        rotation: PropTypes.arrayOf(PropTypes.number),
+        rotation: PropTypes.shape({}),
```
shammamah: Is there a particular reason this was changed?

mj3cheun: Yes the rotation prop is a float32array not an array which caused issues when I tried using newer versions of React which are more more strict about proptypes. Therefore I changed the proptype to shape which is a general object.

shammamah: Good catch! Is this the most specific it can get?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [0.0.12] - 2019-05-24
+* Fixed issue with Speck not responding to preset changes
```
shammamah: Put this under A `### Fixed` heading, please!

shammamah: (take a look at the rest of our CHANGELOG to see what it should look like)

mj3cheun: whoops fixed

shammamah: Also please make sure these have periods at the end, like the rest of the CHANGELOG entries. (Small detail, but it's good to be as consistent as possible!)

shammamah:
Please also publish a `0.0.12` version (update version number in `package.json`). I can publish to npm and PyPI!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dash-bio",
-  "version": "0.0.11",
+  "version": "0.0.12",
```
mkcor:
```suggestion
  "version": "0.0.12rc1",
```

shammamah: @mkcor We won't be making any other changes to the package until the 0.1.0 release (where utils have been separated into a different package). I think it's ok to have this as the 0.0.12 release.

mkcor: Oh, I get it! Thanks for clarifying. Therefore, we must ship this PR before #361 -- I'll resolve the file conflicts to make it approvable/mergeable then.

mkcor: Well, we're currently at version 0.1.0rc3 so there's a problem... I guess we need to make this changeset the increment between 0.1.0rc3 and 0.1.0rc4.

mkcor:
Oh, I don't think I have permission to upload version 0.1.0rc4 to https://pypi.org/project/dash-bio/... So I'll keep version number artificially at 0.1.0rc3 and let @shammamah do the package update/upload.

shammamah:
@mkcor 0.1.0 has already been published to both npm and pypi. We'll change this to a 0.1.1 rc version instead after we get #361 merged in.

mkcor:
> @mkcor 0.1.0 has already been published to both npm and pypi. We'll change this to a 0.1.1 rc version instead after we get #361 merged in.

Oh... I see. I have then reviewed #361 (rebuild package to reflect update in `package-json.lock`).

src/lib/components/Speck.react.js
mj3cheun: I don't believe that there is a specific float32array proptype so yes this is the most specific I can get without defining a custom proptype

PR #351, opened 2019-05-12
Initial build of R package to implement dash-bio functionality for DashR

rpkyle:
Closes #350.

## About
- [x] This is a new component
- [x] I am adding a feature to an existing component, or improving an existing feature
- [x] I am closing an issue

## Description of changes

This PR proposes to commit an initial, working build of the `dash-bio` library for use with DashR. Additionally, minor edits to `package.json` have been made to provide additional fields for recording information used in the R package `DESCRIPTION` file.

A parser library for `dashBio` is not included with this release, but may be provided in subsequent versions.

The syntax for `dash-generate-components` used to build this package was:

`dash-generate-components ./src/lib/components dash_bio --r-prefix="dashbio" --r-depends="manhattanly"`

PR #359, opened 2019-05-22
Add docstrings

shammamah:
Related to #351.

### DIFF
--- a/dash_bio/NeedlePlot.py
+++ b/dash_bio/NeedlePlot.py
@@ -5,7 +5,7 @@ from dash.development.base_component import Component, _explicitize_args
 
 class NeedlePlot(Component):
     """A NeedlePlot component.
-The Needle Plot component is used to display large datasets
+The Needle Plot component is used to visualize large datasets
 containing categorical or numerical data. The lines and markers in
 the plot correspond to bars in a histogram.
 
@@ -13,7 +13,7 @@ Keyword arguments:
 - id (string; optional): The ID of this component, used to identify dash components
 in callbacks. The ID needs to be unique across all of the
 components in an app.
-- mutationData (optional): The data that are displayed on the plot.. mutationData has the following type: dict containing keys 'x', 'y', 'mutationGroups', 'domains'.
+- mutationData (optional): The data that are displayed on the plot. mutationData has the following type: dict containing keys 'x', 'y', 'mutationGroups', 'domains'.
 Those keys have the following types:
   - x (string | list; optional)
   - y (string | list; optional)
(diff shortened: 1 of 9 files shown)


dash_bio/NeedlePlot.py
```diff
@@ -5,30 +5,32 @@
 
 class NeedlePlot(Component):
     """A NeedlePlot component.
-
+The Needle Plot component is used to display large datasets
```
rpkyle: @shammamah I wonder if it would make more sense to write "used to **visualize** large datasets" rather than "_display_ large datasets" here?

dash_bio/NeedlePlot.py
```diff
@@ -5,30 +5,32 @@
...
+containing categorical or numerical data. The lines and markers in
+the plot correspond to bars in a histogram.
 
 Keyword arguments:
 - id (string; optional): The ID of this component, used to identify dash components
 in callbacks. The ID needs to be unique across all of the
 components in an app.
-- mutationData (optional): . mutationData has the following type: dict containing keys 'x', 'y', 'mutationGroups', 'domains'.
+- mutationData (optional): The data that are displayed on the plot.. mutationData has the following type: dict containing keys 'x', 'y', 'mutationGroups', 'domains'.
```
rpkyle: Looks like an extra `.` after the word plot here.

dash_bio/NeedlePlot.py
```diff
@@ -5,30 +5,32 @@
...
   - y (string | list; optional)
   - mutationGroups (list; optional)
   - domains (list; optional)
-- xlabel (string; optional)
-- ylabel (string; optional)
-- rangeSlider (boolean; optional)
-- needleStyle (optional): . needleStyle has the following type: dict containing keys 'stemColor', 'stemThickness', 'stemConstHeight', 'headSize', 'headColor', 'headSymbol'.
+- xlabel (string; optional): Title of the x-axis.
+- ylabel (string; optional): Title of the y-axis.
+- rangeSlider (boolean; optional): If True, enables a rangeslider for the x-axis.
```
rpkyle: Whenever describing a boolean in docstrings, it would help to use sentence case, I think. When I see `True` instead of `true` I immediately think Python, and these are the same docstrings that will get inserted into R (and any other language) documentation too 🤓

dash_bio/NeedlePlot.py
```diff
@@ -5,30 +5,32 @@
...
+- needleStyle (optional): Options for the needle marking single site mutations.. needleStyle has the following type: dict containing keys 'stemColor', 'stemThickness', 'stemConstHeight', 'headSize', 'headColor', 'headSymbol'.
```
rpkyle: That pesky double period strikes again!

dash_bio/NeedlePlot.py
```diff
@@ -5,30 +5,32 @@
...
 Those keys have the following types:
   - stemColor (string; optional)
   - stemThickness (number; optional)
   - stemConstHeight (boolean; optional)
   - headSize (number; optional)
   - headColor (list | string; optional)
   - headSymbol (list | string; optional)
-- domainStyle (optional): . domainStyle has the following type: dict containing keys 'domainColor', 'displayMinorDomains'.
+- domainStyle (optional): Options for the protein domain coloring.. domainStyle has the following type: dict containing keys 'domainColor', 'displayMinorDomains'.
```
rpkyle: ... nice try, pesky period, but I see you!

dash_bio/NeedlePlot.py
shammamah: Ugh, this looks like it's a problem with the Dash class generator. Period only shows up after the prop description if it's an object. I'll fix this just to get it to work now, but will need to make a PR in the Dash repo to fix the autogeneration.

rpkyle: I sympathize, this has been the week of `dash-generate-components` for me.

rpkyle:
:dancer:

PR #363, opened 2019-05-27
Add mol2d

shammamah:
## About
- [x] This is a new component
- [ ] I am closing an issue

## Description of changes
* Added Molecule2dViewer component. 
* Added Molecule2dViewer demo application.
* Merged in changes from #355 and added those in to be part of the `0.1.1` release. 

## Before merging


dash_bio/metadata.json
```diff
@@ -1763,6 +1763,135 @@
...
+        },
+        "required": false,
+        "description": "The ID used to identify this component in callbacks"
+      },
+      "setProps": {
+        "type": {
+          "name": "func"
+        },
+        "required": false,
+        "description": "Dash-assigned callback that should be called whenever properties change"
```
mkcor: This description should end with a period to be consistent with the following ones (below). Regarding its content, I'm not sure about writing the "that should be" part... If I understand correctly, this callback gets called whenever properties change, I wouldn't expect the end-user to have to do it 'manually' ("should be" sounds like an instruction in this context). Please enlighten me :)

dash_bio/metadata.json
```diff
@@ -1763,6 +1763,135 @@
...
+        "returns": null
+      }
+    ],
+    "props": {
+      "id": {
+        "type": {
+          "name": "string"
+        },
+        "required": false,
+        "description": "The ID used to identify this component in callbacks"
```
mkcor:
```suggestion
        "description": "The ID used to identify this component in callbacks."
```

mkcor: Oh, please transpose my comments to the source (`src/lib/components/Molecule2dViewer.react.js`).

src/lib/index.js
```diff
@@ -8,6 +8,7 @@ import Circos from './components/Circos.react';
 import OncoPrint from './components/OncoPrint.react';
 import AlignmentChart from './components/AlignmentChart.react';
 import AlignmentViewer from './components/AlignmentViewer.react';
+import Molecule2dViewer from './components/Molecule2dViewer.react';
```
mkcor: I thought we once reordered these imports alphabetically... As the number of components increase, it will be helpful to read them in alphabetical order. Could you please reorder both imports and exports?

tests/dashbio_demos/app_molecule2d.py
```diff
@@ -0,0 +1,230 @@
...
+
+# running with gunicorn (on servers)
+elif 'DASH_PATH_ROUTING' in os.environ:
+    from tests.dashbio_demos.utils.app_standalone import run_standalone_app
+
+DATAPATH = os.path.join(".", "tests", "dashbio_demos", "sample_data", "mol2d_")
+
+
+def description():
+    return 'Two-dimensional visualizations of molecular structure.'
```
mkcor:
```suggestion
    return 'Two-dimensional visualization of molecular structures.'
```

tests/dashbio_demos/app_molecule2d.py
```diff
@@ -0,0 +1,230 @@
...
+                                    title='Change the base bond length',
+                                    className='app-controls-block',
+                                    children=[
+                                        html.Div(className='app-controls-name',
+                                                 children='Bond length'),
+                                        dcc.Slider(
+                                            id='mol2d-bond-length',
+                                            min=5,
+                                            max=100,
+                                            value=20
```
mkcor: I think these numbers should be displayed in the UI. Is their unit nanometer, picometer? It's not clear to me what "base bond length" means in this context; it sounds like something with chemical reality, but it looks like a choice in representation / visualization / rendering... So I would expect it to have a name more like "unit bond length" or "bond length coefficient"...

mkcor:
Great addition to the Dash Bio gallery and 'catalog' of components!!
In light of the Ideogram test fails, I have tried setting all remaining `wait_for_elements_by_css_selector()` calls (after element update) to a timeout of 20 (I should have tried 50)... to no clear avail.

tests/dashbio_demos/app_molecule2d.py
mkcor: Actually, playing with the demo app, I can see that going below or beyond certain (threshold?) values, the molecule structure completely changes... So, what's the chemical reality behind this "base bond length?" Also, I find the representation pretty 3D-looking, not 2D...

dash_bio/metadata.json
shammamah: This is just added from the cookiecutter boilerplate (which was used to generate this repository), and is consequently part of all of the components: https://github.com/plotly/dash-component-boilerplate/blob/64406a8ecec85da56b40e930d48fa66d1a3b309c/%7B%7Bcookiecutter.project_shortname%7D%7D/src/lib/components/%7B%7Bcookiecutter.component_name%7D%7D.react.js#L40

`setProps` is actually not necessarily called every time properties change; if you look in the code for the components, the author _does_ have to "manually" add in a call to `setProps`. Using this function allows for exchange of information between the React component and the Dash app.

I agree that it could be worded better (e.g., "Dash-assigned callback that should be called whenever information regarding property changes needs to be exchanged with Dash."), but "should be called" is correct here as far as I can tell.

src/lib/index.js
shammamah: Fixed in fa6f132 and 8add7f7!

dash_bio/metadata.json
shammamah: Fixed in 7bb2b00!

tests/dashbio_demos/app_molecule2d.py
shammamah: The reason I didn't display the numbers is precisely _because_ there are no units specified. The distance is calculated here as a function of the *x* and *y* coordinates given in the PubChem file for the molecule: https://github.com/plotly/dash-bio-utils/blob/bea1979a9990f729248875852f44b57b82b3a291/dash_bio_utils/chem_structure_reader.py#L18

The PubChem files look like this, and the coordinates are retrieved from the "coords" part: https://pubchem.ncbi.nlm.nih.gov/rest/pug/compound/CID/2244/record/JSON/?record_type=2d&response_type=display

However, I found that the structure of the molecule does change quite a bit (like you said) and isn't accurate if the multiplier is just 1 (i.e., if there is no multiplier at all). I could change it to "bond length coefficient/multiplier", since I think that makes more sense.

Regarding your comment about 3D - the component that I wrapped just has the name `Molecule2d`, so I'm not sure what to tell you. All of the data are also taken from PubChem 2d structural data (see the file for aspirin that I linked above -- there are only x and y coordinates, so it is technically a flat diagram).

tests/dashbio_demos/app_molecule2d.py
shammamah: Fixed in a23cc81!

mkcor: Since a23cc81 doesn't contain any other change and has a non-compliant commit message, it would have been better (in this case) to commit via the interface the lazy way...

tests/dashbio_demos/app_molecule2d.py
mkcor: > Regarding your comment about 3D - the component that I wrapped just has the name Molecule2d, so I'm not sure what to tell you. All of the data are also taken from PubChem 2d structural data (see the file for aspirin that I linked above -- there are only x and y coordinates, so it is technically a flat diagram).

Thanks for the references. Arguments of authority are good to be aware of, but it's even better to have some minimal understanding of the objects we manipulate... In this moment, I wish I did as much chemistry as most physicists do in their training! The 2D representations of chemical molecules I have in mind are those: https://en.wikipedia.org/wiki/Skeletal_formula

I guess it boils down to the notion of *embedding*: 2D vs 3D don't refer, here (with `Molecule3dViewer` and `Molecule2dViewer`), to the molecular structure (which is 3D in both cases) but to the space where the structure is embedded. With `Molecule2dViewer`, we have a 2D drawing of a 3D shape.

dash_bio/metadata.json
mkcor: Ok, thanks for the details and for pointing me to the source! I'll poke @Marc-Andre-Rivet on this then ;)

tests/dashbio_demos/app_molecule2d.py
mkcor: > I could change it to "bond length coefficient/multiplier", since I think that makes more sense.

Yes, it is definitely a coefficient/multiplier, so I think it would be clearer to use this name in the UI (plus, a random user doesn't have access to the [docstring](https://github.com/plotly/dash-bio-utils/blob/bea1979a9990f729248875852f44b57b82b3a291/dash_bio_utils/chem_structure_reader.py#L30)).

> However, I found that the structure of the molecule does change quite a bit (like you said) and isn't accurate if the multiplier is just 1 (i.e., if there is no multiplier at all).

I would like to understand this... when the multiplier is 1, we end up with the Euclidean distance... why does that lead to an inaccurate molecular structure? Why/how is the accurate molecular structure obtained by multiplying the Euclidean distance by a coefficient > 1? Any coefficient > 1?

mkcor: > The reason I didn't display the numbers is precisely because there are no units specified.

Ok, now that I know it's a coefficient, I understand it has no unit.

 > The distance is calculated here as a function of the x and y coordinates given in the PubChem file for the molecule: https://github.com/plotly/dash-bio-utils/blob/bea1979a9990f729248875852f44b57b82b3a291/dash_bio_utils/chem_structure_reader.py#L18

> The PubChem files look like this, and the coordinates are retrieved from the "coords" part: https://pubchem.ncbi.nlm.nih.gov/rest/pug/compound/CID/2244/record/JSON/?record_type=2d&response_type=display

It's now clear to me how the data in "coords" is used to compute the Euclidean distance, but what are the rules/reasons for multiplying this Euclidean distance by a coefficient > 1? Is there a reference in the chemistry literature explaining that scaling the Euclidean distance between atoms yields an 'accurate' molecular structure (or conventional representation thereof?)?

shammamah: It's due to the bond strength. The component calculates the structure based on the bond length and the bond strengthm, which are just used as "forces": 

https://github.com/Autodesk/molecule-2d-for-react/blob/d0030eb0521032cbaa9f20566e8db9166737647a/src/components/molecule_2d.jsx#L122
https://github.com/d3/d3/blob/master/API.md#forces-d3-force

So it makes sense that if the strength is left unchanged when the bond distance changes, it leads to an imbalance of forces and the structure conforms itself in a way that balances them.

shammamah: I've changed the name of the control, and added more clarification to the text underneath it, in a1ece88.

mkcor: Nice change (a1ece88) and I think it would be more than fair to show the 1 and 10 (or, apparently, 10 and 100) mark on the slider.

mkcor: > It's due to the bond strength.

I understand the model, but what reality does it (is it supposed to) reflect? When does bond length change? When temperature changes? When the molecule finds itself in a different solvent?

For instance, when using the slider to increase the bond length multiplier, are we mimicking the effect of a temperature increase?

I'm familiar with https://en.wikipedia.org/wiki/Bond_energy and then the picture is that for a given molecule (say, methane `H(C–H)`) at a given temperature (say, 298 K), a given molecular bond (say, the `(C–H)` bond) has a certain strength corresponding to a certain length... and I'm not aware of the phenomena that scale up that length (which we model in our visualization). What are they?

shammamah: I don't think this component is meant to be an accurate simulation, but rather just a rendering tool. I would imagine that if it were used in a simulation, the bond distances and strengths would be calculated somewhere else and merely supplied to the component. It's up to the user what the bond length and bond strength mean. 

I can't seem to find any relevant documentation related to the bonds in the `molecule-2d-for-react` repository that makes reference to some real chemical/physical quantity. It might be a good idea to open up an issue in that repo and ask them what it means.

shammamah: I'm personally against adding numbers to the bond length multiplier control, because (as we established) it doesn't really mean anything. It isn't useful information to the user whether the bond length is multiplied by 10 or 100, since there isn't really a (chemically) meaningful quantitative property related to the bond length in the first place.

mkcor: Of course, this is not a simulation tool, only a rendering one. But I would like to have a *tiny* clue about what it is that we're representing...

> It's up to the user what the bond length and bond strength mean.

Well, no... the bond length and bond strength *do* mean something real and, apparently, they can vary (can they? if needed, I'll look for references tomorrow) and if we feature this in the Dash Bio gallery, we want to be able to tell a super basic message such as "that's a 2D representation for 3D molecule structures where bond length can be varied, which corresponds to simulating XXX" and *minimally* we need a word or a couple words for XXX, even if very vague (e.g., "temperature variation") to express *qualitatively*, very *vaguely*, what it is we're representing, even with zero scientific rigor.

> I can't seem to find any relevant documentation related to the bonds in the molecule-2d-for-react repository that makes reference to some real chemical/physical quantity. It might be a good idea to open up an issue in that repo and ask them what it means.

Definitely!

> (as we established)

As far as I know, we haven't established anything...

> there isn't really a (chemically) meaningful quantitative property related to the bond length in the first place

I would love to believe you... do you have a reference for this claim? We need a source, some chemistry publication, that describes *minimally* how and why bond length multipliers > 1 are used conventionally in 2D representations of 3D molecular structures. I can ask my chemistry major friends tomorrow.

mkcor: Ok, I haven't grabbed a textbook to review chemistry 101 yet, but I've had the privilege of chatting with my friend Hélène Seiler. She is a PhD in physical chemistry and she works on world-class experiments investigating precisely the stretching, bending, breaking of chemical bonds!

The molecular structure corresponding to the given "coords" (from PubChem) does correspond to an existing reality (it's the structure at equilibrium). I think it would be most natural to showcase this 'default' case (multiplier = 1), so that at least we know this representation means something (molecular structure at equilibrium).

In our demo app, this 'default' case (multiplier = 1) becomes a 'starting point' for a process modelled by the linear increase of bond lengths: I guess we can view it as a specific case of bond stretching. We don't explain this process; I'm fine saying this is not our job; we humbly provide a visualization tool rendering what could be obtained from certain computations, simulations (of which meaning/relevance we don't discuss).

In light of this, I suggest starting the multiplier slider at 1 and changing the instruction "Change the bond length multiplier. This will cause the structure of the molecule to change as the atoms are pushed farther apart." into:

"Increase bond lengths linearly from their values at equilibrium. This visualization will be reminiscent of chemical bond stretching."

shammamah: This change has been made in 7d165ad.

mkcor:
As I double-checked the deployment, I noticed a line break between "Bond length" and "multiplier" (it looks like it should fit on one line, just like "Search for molecule by name"):
line_break

shammamah:
@mkcor What browser are you using? It looks fine to me on Vivaldi, Chrome, and Safari.

mkcor:
> @mkcor What browser are you using? It looks fine to me on Vivaldi, Chrome, and Safari.

Firefox!

PR #364, opened 2019-05-28
Double timeout for Ideogram tests which fail often.

mkcor:
## About
The `selenium` utility https://github.com/plotly/pytest-dash/blob/d0291909b1230f3c39473b6f038ae7a6b48670c3/pytest_dash/wait_for.py#L44 often fails to find the elements identified in some of our tests (specifically in file `tests/test_ideogram.py`) within the default timeout. Values we test for are therefore not updated, leading to CI build failures (we then relaunch CI builds manually).

## Description of changes
This changeset won't change the world, but it could spare us a few CI fails (iirc, last time I toyed with this, I ended up not submitting the change because I wasn't able to observe these timeout-related fails locally). Also, note that this change will be subject to the upcoming refactoring of `pytest-dash` by @byronz.

## Before merging

### DIFF
--- a/tests/test_ideogram.py
+++ b/tests/test_ideogram.py
@@ -234,7 +234,7 @@ def test_ploidy(dash_threaded):
     btn.click()
 
     # assert doubling of the 22 chromosomes + X and Y chromosomes
-    chromosomes = wait_for_elements_by_css_selector(driver, '.chromosome')
+    chromosomes = wait_for_elements_by_css_selector(driver, '.chromosome', timeout=20)
     assert len(chromosomes) == 46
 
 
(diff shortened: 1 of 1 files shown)


byronz:
> 💃 Let's hope that this works until we can get a better fix with `pytest-dash`!

you mean a better fix with `dash`?

mkcor:
@shammamah @byronz this one is not really a fix, again, it's just that we (selenium) will be *more likely* to find those elements using CSS... that tended to take more than the default timeout. I don't expect the `pytest-dash` refactoring to improve this (I expect this refactoring to change essentially where we import those testing utilities from)... I think the 'limiting factor' here is our logic of looking for elements with the CSS selector (to test for expected updates in the UI). I haven't spent time digging deeper how 'good' integration tests should be done...

PR #344, opened 2019-05-03
Update some JS dependencies and rebuild package.

mkcor:
## About
This is about managing JS dependencies.

## Description of changes
My original intention was to migrate to React 16 and Dash 0.42, but I'll tackle it next week. For now, I have only run:
```
npm install
npm audit fix
npm run build:all
```

I'm not sure what is going on with the following:
* is Renote Bot still running? (We haven't seen any of its automated PR in a while...)
* although we haven't migrated to React 16 as a dependency, we are using it as a dependency of dependency(??) https://github.com/plotly/dash-bio/blob/4c12d9206ba430c7c8ea431b116c9d035034b513/package-lock.json#L10775
* I have browsed the [docs](https://docs.npmjs.com/auditing-package-dependencies-for-security-vulnerabilities#security-vulnerabilities-found-requiring-manual-review) but haven't actually handled security vulnerabilities *manually* so far, and my understanding is that a `package-lock.json` file shouldn't be edited manually anyway... All this to say that I haven't addressed https://github.com/plotly/dash-bio/network/alert/dash_bio/package-lock.json/jquery/open in this PR.

fyi @Marc-Andre-Rivet 

## Before merging
- [ ] I have gone through the [code review checklist](https://github.com/plotly/dash-component-boilerplate/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/review_checklist.md)
- [ ] I have completed all of the [steps to take before merging](https://github.com/plotly/dash-bio/blob/master/.github/before_merging.md)
- [ ] I know how to [deploy to DDS](https://github.com/plotly/dash-bio/blob/master/.github/deploying_to_dds.md)

PR #367, opened 2019-05-31
Clean up dependency declarations.

mkcor:
## About
This is about dependency management.

## Description of changes
I was about to start a WIP branch where I would use `dash-bio-utils==0.0.3rc1`. Since version `0.0.2` is available, we might as well use it; meanwhile I fixed a couple other things.

## Before merging

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
     long_description=long_description,
     long_description_content_type='text/markdown',
     install_requires=[
-        'dash>=0.40.0',
+        'dash==0.40.0',
         'pandas',
         'plotly',
         'scipy',
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@
 cython>=0.19
 dash==0.40.0
 dash-bio==0.1.2-rc1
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -24,7 +24,7 @@
     long_description=long_description,
     long_description_content_type='text/markdown',
     install_requires=[
-        'dash>=0.40.0',
+        'dash==0.40.0',
```
shammamah: Why are we locking it to this version?

mkcor: For the same reason we've been holding off migrating to React 16 + Dash 0.42, i.e., tests don't pass, even if the actual apps work well:

* https://github.com/plotly/dash-bio/pull/344#issuecomment-490018688
* https://github.com/plotly/pytest-dash/pull/76

We locked this dependency a while ago:

* https://github.com/plotly/dash-bio/commit/7d4a1a238a2bc5fe604b5081d5846762e4bacf79
* https://github.com/plotly/dash-bio/commit/d093651162a1fb3a5103f9ea4866088c9dc55002

We'll be able to move forward as @byronz refactors `pytest-dash`.

Issue #330, opened 2019-04-25
Several utility functions use an argument pattern where both options (file or data) default to an empty value.


mkcor:
This looks weird to me. It looks like we use this function by specifying either `contents` or `filepath`, is that correct? Then, I understand why the other argument doesn't matter and comes with a meaningless default value... But I don't find it very clean. Let's take sides and favour either `contents` or `filepath`.

_Originally posted by @mkcor in https://github.com/plotly/dash-bio/pull/327_

Let's determine whether that's common practice or not, but I doubt since a syntax such as
```py
def read_fasta(
        file_path='',
        data_string='',
)
```
implies you could call `read_fasta()` without specifying arguments, since they come with default values... but really `read_fasta()` wouldn't make sense (it's ill-defined, so it should error, but I don't see any error handling in the function definition https://github.com/plotly/dash-bio/blob/96d43e90c4598b42282ee27b28e827c8e0f6ec7e/dash_bio/utils/protein_reader.py#L29).

mkcor:
Closing this issue, because utilities now live in their own package, `dash_bio_utils`.

PR #375, opened 2019-06-18
Update dash-bio-utils dependency (standardized data parsers).

mkcor:
Directly related to issue #330.

## About
- [ ] This is a new component
- [ ] I am closing an issue

## Description of changes
In these changes, we are making use of updated data parsers (utils, think `read_*`), which have a more standard syntax (inspiration: Pandas) and more consistency with one another (e.g., `read_soft_file` is now `read_soft`).

## Before merging


mkcor:
Thank you so much for assisting with this, @shammamah!

Question: Why is the missing newline at EOF so desirable in `package-info.json` file? I'm referring to commit 87537be.

shammamah:
No idea. That's just how the file is generated. I can add one in here: https://github.com/plotly/dash-bio/blob/master/get_version_info.py

mkcor:
@shammamah on a side note: The fact that commit 8e341e7 ("Update read_structure() and read_xyz() calls") passed the tests, while `read_soft_file()` hadn't been updated/adapted yet, shows that our test coverage is poor. I guess `pylint` would have caught an issue if we imported the function directly, but we don't (we import the module where it lives)...

mkcor:
> No idea. That's just how the file is generated. I can add one in here: https://github.com/plotly/dash-bio/blob/master/get_version_info.py

I would say so.

Ok, so 87537be doesn't have any particular meaning -- why did you commit/push that change then?

shammamah:
It was automatic -- I suppose I'm just used to adding the autogenerated files whenever I re-generate the bundle. I can revert that particular commit if necessary.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Changed
 * Made Speck component compatible with refactored Speck library.
 * Optimized Speck component further to increase performance.
+* Update dash_bio_utils where data parsers were standardized.
```
shammamah:
```suggestion
* Updated dash_bio_utils to latest version, in which data parsers were standardized.
```

mkcor: Right, I forgot about the preterite!

I'm not sure "latest version" is very fitting, because:

* strictly speaking, in a changelog context, it should be "then-latest";
* the "version" in question is a "release candidate" so even knowing it was the "then-latest version" by cross-matching release dates, it wouldn't be accurate;
* the point of the change is not to update to the latest version of utils (say, just because it's available, which would totally make sense, but it's just not the point here), it's to take advantage of these more standard data parsers.

If you suggest "in which..." why not "wherein" at this point? ;)

mkcor:
> It was automatic -- I suppose I'm just used to adding the autogenerated files whenever I re-generate the bundle. I can revert that particular commit if necessary.

It's not a big deal... it depends on when we update this file generator (in terms of avoiding or adding meaningless diff's and, hence, commits).

I tend to be against doing things in an 'automatic' way: I find that doing
```bash
git status
git diff
git add <files I really mean or even lines I mean (interactively)>
git status
git add <possibly other files or lines>
git status
git commit -m "Meant to add that changeset"
```
(basically running `git status` all the time) is a good workflow for building a meaningful commit log (anecdotally, I have taught [Git](http://swcarpentry.github.io/git-novice/) many times, sometimes not amazingly, which has taught me a lot in return).

mkcor:
Alternatively, we could have a standard procedure for committing auto-generated / auto-edited files (together or separately? following a commit message template?).

Speaking of which... hehe I don't know where the auto-edition of `package-info.json` comes from... When I ran `npm run build:all`, it changed `bundle.js` but not `package-info.json`. Could you please enlighten me? :)

CHANGELOG.md
shammamah: >I'm not sure "latest version" is very fitting, because: ... 

Those are all valid points. Maybe instead it could say something like "Updated dash_bio_utils to version that introduced standardization to data parsers." 

>If you suggest "in which..." why not "wherein" at this point? ;)

Haha, we can have that too! Fewer characters to change :P

shammamah:
>Alternatively, we could have a standard procedure for committing auto-generated / auto-edited files (together or separately? following a commit message template?).

Together makes the most sense to me -- what do you think? I've mostly been doing "Add autogenerated files", or some version thereof.

>Could you please enlighten me? :)

Sure! The script `get_version_info.py` handles all of that. It is called here in the build script: https://github.com/plotly/dash-bio/blob/master/package.json#L21

Here is the script: https://github.com/plotly/dash-bio/blob/master/get_version_info.py

>When I ran npm run build:all, it changed bundle.js but not package-info.json.

This worries me, since it should have done the same thing for you, too... :/ 

(Side note: I love "Wolfman and Dracula have been hired by Universal Missions (a space services spinoff from Euphoric State University) to investigate if it is possible to send their next planetary lander to Mars.")

CHANGELOG.md
mkcor: Great, thanks, I'll go for "Updated dash_bio_utils to version that introduced standardization of data parsers" (because we didn't introduce standardization *to* data parsers so much, we introduced standardized data parsers, we introduced some standardization of our data parsers).

mkcor:
> > [...] standard procedure for committing auto-generated / auto-edited files (together or separately? following a commit message template?).
> 
> Together makes the most sense to me -- what do you think? I've mostly been doing "Add autogenerated files", or some version thereof.

I agree. Committing these autogenerated changes together makes sense to me.

So (careful!) I'll rebase and remove the isolated edition of `package-info.json`.

> (Edit: It's because the version wasn't updated in `package.json` before the bundle was generated (i.e., before `npm run build:all` was run) -- [34515ac](https://github.com/plotly/dash-bio/commit/34515ac4e9314a4928dd13d1885d59ca08824aa9) comes before [038c0c4](https://github.com/plotly/dash-bio/commit/038c0c4fbe90c7e1e3bd9d2ecf597fe578b315e9) in the commit history for this PR. The `get_version_info` script reads `package.json` to write the version info.)

Oh, thank you! I have to be more rigorous with this workflow.
 
> (Side note: I love "Wolfman and Dracula have been hired by Universal Missions (a space services spinoff from Euphoric State University) to investigate if it is possible to send their next planetary lander to Mars.")

I know! ;)

mkcor:
@shammamah Should be ready for final review now! Thanks.

PR #379, opened 2019-06-25
Python docstrings refactoring

### LINKED ISSUE #378: Dash Bio Python components require docstrings formatted like those from other libraries
As of https://github.com/plotly/dash/pull/782, the docstrings for components' props now look like this: 

`[prop name] ([type]; [required/optional/default [default value]]): [prop description]`

The `Clustergram`, `ManhattanPlot`, and `VolcanoPlot` docstrings should be fixed accordingly.

shammamah:
Closes #378

## About
- [ ] This is a new component
- [x] I am closing an issue

## Description of changes
- move all information about default values to the beginning of the docstring, between the parentheses; if a default value is defined, replace "optional" with "default [default value]"
- change "int" and "float" to "number"

## Before merging


dash_bio/component_factory/_clustergram.py
```diff
@@ -50,15 +50,14 @@ def Clustergram(
...
 
 - data (ndarray; required): Matrix of observations as array of arrays
-- generate_curves_dict (bool; optional): Whether or not to return a
+- generate_curves_dict (bool; default false): Whether or not to return a
     dictionary containing information about the cluster number
     associated with each curve number in the graph. (May be useful
     if one wishes to capture the cluster number that is clicked.)
-    (Default: False)
-- return_computed_traces (bool; optional): Whether or not to return
+- return_computed_traces (bool; default false): Whether or not to return
```
```suggestion
- return_computed_traces (bool; default False): Whether or not to return
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -50,15 +50,14 @@ def Clustergram(
 Keyword arguments:
 
 - data (ndarray; required): Matrix of observations as array of arrays
-- generate_curves_dict (bool; optional): Whether or not to return a
+- generate_curves_dict (bool; default false): Whether or not to return a
```
```suggestion
- generate_curves_dict (bool; default False): Whether or not to return a
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -50,15 +50,14 @@ def Clustergram(
...
     dictionary containing information about the cluster number
     associated with each curve number in the graph. (May be useful
     if one wishes to capture the cluster number that is clicked.)
```
```suggestion
    for capturing the cluster number that is clicked.)
```
I know it's not part of the PR, just noticed while reviewing...

dash_bio/component_factory/_clustergram.py
```diff
@@ -67,66 +66,65 @@ def Clustergram(
...
+- dist_fun (function; default scipy.spatial.distance.pdist): Function
+    to compute the pairwise distance from the observations (see
+    scipy.spatial.distance.pdist).
+- link_fun (function; default scipy.cluster.hierarchy.linkage): Function to
+    compute the linkage matrix from the pairwise distances (see
+    scipy.cluster.hierarchy.linkage).
+- color_threshold (dict; default {'row': 0, 'col': 0}): Maximum
+    linkage value for which unique colors are assigned to clusters;
+    'row' for rows, and 'col' for columns.
+- optimal_leaf_order (bool; default false): Enabling/disabling of the
```
```suggestion
- optimal_leaf_order (bool; default False): Enabling/disabling of the
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -67,66 +66,65 @@ def Clustergram(
...
-    percentile, would be gray; and the cell(s) in the 100th percentile
-    would be colored black. (Default: [[0.0, 'rgb(255,0,0)'], [0.5,
-    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']])
+    neighboring leaves.
+- color_map (list; default [[0.0, 'rgb(255,0,0)'], [0.5,
+    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']]): The colorscale for the
+    heatmap. Each list element contains two elements; the first
+    element refers to the portion of the maximum data point under
+    which a cell will be colored, and the second element refers to the
+    color. e.g., a colorscale [[0.0, 'white'], [0.5, 'gray'], [1.0,
```
```suggestion
    color, e.g., a colorscale [[0.0, 'white'], [0.5, 'gray'], [1.0,
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -67,66 +66,65 @@ def Clustergram(
...
+    'black']] mean that for all cells with a value in the 50th or
```
```suggestion
    'black']] means that for all cells with a value in the 50th or
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -67,66 +66,65 @@ def Clustergram(
...
+    lower percentile of the dataset, the color on the heatmap would be
+    white; all cells with a value in the 50th or higher percentile,
```
mkcor: There is a degeneracy issue here (speaking like a physicist): Are cells with a value in the 50th percentile white or gray?
We have either (`<=` and `>`) or (`<` and `>=`), not (`<=` and `>=`).

dash_bio/component_factory/_clustergram.py
```diff
@@ -67,66 +66,65 @@ def Clustergram(
...
     will be colored with one shade, and the values that are above this
-    value will be colored with another. (Default: 3)
-- symmetric_value (bool; optional): Whether or not to center the
-    values of the heatmap about zero. (Default: True)
-- log_transform (bool; optional): Whether or not to transforms the
-    data by taking the base-two logarithm of all values in the
-    dataset. (Default: False)
-- display_ratio (list/float; optional): The dendrograms' heights with
+    value will be colored with another.
+- symmetric_value (bool; default true): Whether or not to center the
```
```suggestion
- symmetric_value (bool; default True): Whether or not to center the
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -67,66 +66,65 @@ def Clustergram(
...
+    values of the heatmap about zero.
+- log_transform (bool; default false): Whether or not to transforms
```
```suggestion
- log_transform (bool; default False): Whether or not to transform
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -136,38 +134,36 @@ def Clustergram(
...
-   None)
+   https://plot.ly/python/reference/#bar-marker-colorbar-tickfont).
 - annotation_font (dict; optional): The font options for annotations,
    as specified in the Plotly graph_objs documentation (see:
-   https://plot.ly/python/reference/#layout-scene-annotations-items-annotation-font). (Default:
-   None)
-- line_width (list/float; optional): The line width for the
+   https://plot.ly/python/reference/#layout-scene-annotations-items-annotation-font).
+- line_width (list | number; default 0.5): The line width for the
     dendrograms. If in list format, the first element corresponds to
```
mkcor: Indentation seems inconsistent... Items above are indented with only one space (within item).

dash_bio/component_factory/_clustergram.py
```diff
@@ -136,38 +134,36 @@ def Clustergram(
...
-    to the width of the column dendrogram traces. (Default: 0.5)
-- paper_bg_color (string; optional): The background color of the paper on the
-    graph. (Default: 'rgba(0,0,0,0)')
-- plot_bg_color (string; optional): The background color of the subplots on the
-    graph. (Default: 'rgba(0,0,0,0)')
-- height (int; optional): The height of the graph, in px. (Default:
-  500)
-- width (int; optional): The width of the graph, in px. (Default: 500)"""
+    to the width of the column dendrogram traces.
+- paper_bg_color (string; default 'rgba(0,0,0,0)`): The background
```
```suggestion
- paper_bg_color (string; default 'rgba(0,0,0,0)'): The background
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -281,39 +274,35 @@ class _Clustergram():
...
-- paper_bg_color (string; optional): The background color of the paper on the
-    graph. (Default: 'rgba(0,0,0,0)')
-- plot_bg_color (string; optional): The background color of the subplots on the
-    graph. (Default: 'rgba(0,0,0,0)')
-- height (int; optional): The height of the graph, in px. (Default:
-  500)
-- width (int; optional): The width of the graph, in px. (Default: 500)"""
-
+    to the width of the column dendrogram traces.
+- paper_bg_color (string; default 'rgba(0,0,0,0)`): The background
```
```suggestion
- paper_bg_color (string; default 'rgba(0,0,0,0)'): The background
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -78,80 +78,76 @@ def ManhattanPlot(
...
             - a numeric quantity to plot such as a p-value or zscore
-- chrm (string; optional): A string denoting the column name for the
-    chromosome.  This column must be float or integer.  Minimum number
-    of chromosomes required is 1. If you have X, Y, or MT chromosomes,
-    be sure to renumber these 23, 24, 25, etc. (Default: "CHR")
-- bp (string; optional): A string denoting the column name for the
-    chromosomal position. (Default: "BP")
-- p (string; optional): A string denoting the column name for the
+- chrm (string; default 'CHR'): A string denoting the column name for
+    the chromosome.  This column must be float or integer.  Minimum
```
mkcor:
```suggestion
    the chromosome. This column must be float or integer. Minimum
```
English style conventions are different in Python (docstrings, comments).

dash_bio/component_factory/_manhattan.py
```diff
@@ -78,80 +78,76 @@ def ManhattanPlot(
...
-    numeric quantity such as peak heights, bayes factors, test
-    statistics. If it is not a p-value, make sure to set logp = FALSE.
-    (Default: "P")
-- snp (string; optional): A string denoting the column name for the
-    SNP names (e.g. rs number). More generally, this column could be
-    anything that identifies each point being plotted. For example, in
-    an Epigenomewide association study (EWAS) this could be the probe
-    name or cg number. This column should be a character. This
+    numeric. This does not have to be a p-value. It can be any numeric
+    quantity such as peak heights, bayes factors, test statistics. If
```
mkcor:
```suggestion
    quantity such as peak heights, Bayes factors, test statistics. If
```
Interesting...

dash_bio/component_factory/_manhattan.py
```diff
@@ -78,80 +78,76 @@ def ManhattanPlot(
...
+    it is not a p-value, make sure to set logp = FALSE.
+- snp (string; default 'SNP'): A string denoting the column name for
+    the SNP names (e.g. rs number). More generally, this column could
```
mkcor:
```suggestion
    the SNP names (e.g., rs number). More generally, this column could
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -78,80 +78,76 @@ def ManhattanPlot(
...
-    graph_objs. (Default: "red")
-- genomewideline_width (number; optional): Width of the genome wide
-  line. (Default: 1)
-- highlight (bool; optional): turning on/off the highlighting of data points
-    considered significant. (Default: True)
-- highlight_color (string; optional): Color of the data points
+    graph_objs.
+- genomewideline_width (number; default 1): Width of the genome wide
+  line.
+- highlight (bool; default true): turning on/off the highlighting of
```
mkcor:
```suggestion
- highlight (bool; default True): turning on/off the highlighting of
```

dash_bio/component_factory/_clustergram.py
shammamah: These are meant to mimic the auto-generated docstrings as closely as possible. "True" and "False" are not capitalized in JavaScript, and they don't end up capitalized in the generated Python docstrings either.

dash_bio/component_factory/_clustergram.py
shammamah: Good catch! This one should stay the same, but the other one ("... 50th or lower...") should be changed.

dash_bio/component_factory/_clustergram.py
shammamah: I'll commit these suggestions and follow up in the Dash repo that is used to generate Python files so that the React components match!

dash_bio/component_factory/_clustergram.py
shammamah: Good catch - fixed in 8c5b90d!

dash_bio/component_factory/_clustergram.py
shammamah: Fixed in 5e83380.

dash_bio/component_factory/_clustergram.py
mkcor: 8c5b90d creates an inconsistency the other way around: items above (lines 150-151 and 153-154) are now indented with three spaces relative to the item (bullet point). You have prepended two spaces instead of one.

dash_bio/component_factory/_manhattan.py
```diff
@@ -78,80 +78,76 @@ def ManhattanPlot(
...
+    The line has no influence on the data points.
+- suggestiveline_color (string; default 'grey'): Color of the suggestive
+  line.
+- suggestiveline_width (number; default 2): Width of the suggestive
+    line.
+- genomewideline_value (bool | float; default -log10(5e-8)): A boolean
+    which must be False to deactivate the option, or a numerical value
+    corresponding to the p-value above which the data points are
+    considered significant.
+- genomewideline_color (string; default 'red'): Color of the genome wide
```
mkcor:
```suggestion
- genomewideline_color (string; default 'red'): Color of the genome-wide
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -78,80 +78,76 @@ def ManhattanPlot(
...
     line. Can be in any color format accepted by plotly_js
-    graph_objs. (Default: "red")
-- genomewideline_width (number; optional): Width of the genome wide
-  line. (Default: 1)
-- highlight (bool; optional): turning on/off the highlighting of data points
-    considered significant. (Default: True)
-- highlight_color (string; optional): Color of the data points
+    graph_objs.
+- genomewideline_width (number; default 1): Width of the genome wide
```
mkcor:
```suggestion
- genomewideline_width (number; default 1): Width of the genome-wide
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -78,80 +78,76 @@ def ManhattanPlot(
...
+  line.
+- highlight (bool; default True): turning on/off the highlighting of
+    data points considered significant.
+- highlight_color (string; default 'red'): Color of the data points
     highlighted because they are significant Can be in any color
```
mkcor:
```suggestion
    highlighted because they are significant. Can be in any color
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -78,80 +78,76 @@ def ManhattanPlot(
...
-    format accepted by plotly_js graph_objs. (Default: "red")
+    format accepted by plotly_js graph_objs.
```
mkcor:
```suggestion
    format accepted by plotly's graph_objs.
```
alternatively, `plotly.graph_objs`, `plotly.js's graph_objs`, ... ?

dash_bio/component_factory/_manhattan.py
```diff
@@ -410,50 +404,47 @@ def figure(
...
+        """Keyword arguments:
+    - title (string; default 'Manhattan Plot') The title of the
+        graph.
+    - showgrid (bool; default true): Boolean indicating whether
+        gridlines should be shown.
+    - xlabel (string; optional): Label of the x axis.
+    - ylabel: (string; default '-log10(p)'): Label of the y axis.
+    - point_size (number; default 5): Size of the points of the
+        Scatter plot.
+    - showlegend (bool; default true): Boolean indicating whether
```
mkcor:
```suggestion
    - showlegend (bool; default True): Boolean indicating whether
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -410,50 +404,47 @@ def figure(
...
-    - ylabel: (string; optional): Label of the y axis. (Default:
-        "-log10(p)")
-    - point_size (number; optional): Size of the points of the Scatter
-        plot. (Default: 5)
-    - showlegend (bool; optional): Boolean indicating whether legends should be
-        shown. (Default: True)
+        """Keyword arguments:
+    - title (string; default 'Manhattan Plot') The title of the
+        graph.
+    - showgrid (bool; default true): Boolean indicating whether
```
mkcor:
```suggestion
    - showgrid (bool; default True): Boolean indicating whether
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -410,50 +404,47 @@ def figure(
...
-        graph_objs. (Default: "red")
-    - genomewideline_width (number; optional): Width of the genome wide
-      line. (Default: 1)
-    - highlight (bool; optional): turning on/off the highlighting of
-        data points considered significant. (Default: True)
-    - highlight_color (string; optional): Color of the data points
-        highlighted because they are significant Can be in any color
-        format accepted by plotly_js graph_objs. (Default: "red")
+        points of the Scatter plot. Can be in any color format
+        accepted by plotly_js graph_objs.
```
mkcor: same question / comment as above

dash_bio/component_factory/_manhattan.py
```diff
@@ -410,50 +404,47 @@ def figure(
...
+    - suggestiveline_value (bool | float; default 8): A value which
+        must be False to deactivate the option, or a numerical value
+        corresponding to the p-value at which the line should be
+        drawn.  The line has no influence on the data points.
```
mkcor:
```suggestion
        drawn. The line has no influence on the data points.
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -410,50 +404,47 @@ def figure(
...
+    - suggestiveline_color (string; default 'grey'): Color of the
+        suggestive line.
+    - suggestiveline_width (number; default 2): Width of the
+        suggestive line.
+    - genomewideline_value (bool/float; default -log10(5e-8)): A
```
mkcor:
```suggestion
    - genomewideline_value (bool | float; default -log10(5e-8)): A
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -410,50 +404,47 @@ def figure(
...
+        boolean which must be False to deactivate the option, or a
+        numerical value corresponding to the p-value above which the
+        data points are considered significant.
+    - genomewideline_color (string; default 'red'): Color of the
+        genome wide line. Can be in any color format accepted by
+        plotly_js graph_objs.
+    - genomewideline_width (number; default 1): Width of the genome
+      wide line.
+    - highlight (bool; default true): turning on/off the highlighting
```
mkcor:
```suggestion
    - highlight (bool; default True): turning on/off the highlighting
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -410,50 +404,47 @@ def figure(
...
+        of data points considered significant.
+    - highlight_color (string; default 'red'): Color of the data
+        points highlighted because they are significant Can be in any
+        color format accepted by plotly_js graph_objs.
 
     Returns:
-    - A figure formatted for plotly.graph_objs."""
+    - A figure formatted for plotly.graph_objs.
```
mkcor: This suggest going for `plotly.graph_objs` version...

dash_bio/component_factory/_manhattan.py
```diff
@@ -410,50 +404,47 @@ def figure(
...
+    - genomewideline_color (string; default 'red'): Color of the
+        genome wide line. Can be in any color format accepted by
+        plotly_js graph_objs.
+    - genomewideline_width (number; default 1): Width of the genome
+      wide line.
+    - highlight (bool; default true): turning on/off the highlighting
+        of data points considered significant.
+    - highlight_color (string; default 'red'): Color of the data
+        points highlighted because they are significant Can be in any
+        color format accepted by plotly_js graph_objs.
```
mkcor:
```suggestion
        color format accepted by plotly.graph_objs.
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -410,50 +404,47 @@ def figure(
...
+        data points are considered significant.
+    - genomewideline_color (string; default 'red'): Color of the
+        genome wide line. Can be in any color format accepted by
+        plotly_js graph_objs.
+    - genomewideline_width (number; default 1): Width of the genome
+      wide line.
+    - highlight (bool; default true): turning on/off the highlighting
+        of data points considered significant.
+    - highlight_color (string; default 'red'): Color of the data
+        points highlighted because they are significant Can be in any
```
mkcor:
```suggestion
        points highlighted because they are significant. Can be in any
```

dash_bio/component_factory/_volcano.py
```diff
@@ -81,70 +81,65 @@ def VolcanoPlot(
...
-    SNP names (e.g. rs number). More generally, this column could be
-    anything that identifies each point being plotted. For example, in
-    an Epigenomewide association study (EWAS) this could be the probe
-    name or cg number. This column should be a character. This
+    statistics. If it is not a p-value, make sure to set logp = FALSE.
+- effect_size (string; default 'EFFECTSIZE'): A string denoting the
+    column name for the effect size. This column in the dataframe must
+    have numeric values, with no values being missing nor NaN.
+- snp (string; default 'SNP'): A string denoting the column name for
+    the SNP names (e.g. rs number). More generally, this column could
```
mkcor:
```suggestion
    the SNP names (e.g., rs number). More generally, this column could
```

dash_bio/component_factory/_volcano.py
```diff
@@ -81,70 +81,65 @@ def VolcanoPlot(
...
+- gene (string; default 'GENE'): A string denoting the column name for
+    the GENE names. More generally, this could be any annotation
+    information that should be included in the plot.
 - annotation (string; optional): A string denoting the column name for
     an annotation. This could be any annotation information that you
     want to include in the plot (e.g. zscore, effect size, minor
-    allele frequency). (Default: None)
-- logp (bool; optional): If True, the -log10 of the p-value is
+    allele frequency).
+- logp (bool; default true): If True, the -log10 of the p-value is
```
mkcor:
```suggestion
- logp (bool; default True): If True, the -log10 of the p-value is
```

dash_bio/component_factory/_volcano.py
```diff
@@ -81,70 +81,65 @@ def VolcanoPlot(
...
+- genomewideline_value (bool | number; default -log10(5e-8)): A
+    boolean which must be False to deactivate the option, or a
+    numerical value corresponding to the p-value above which the data
+    points are considered significant.
+- genomewideline_color (string; default 'red'): Color of the genome
+    wide line. Can be in any color format accepted by plotly_js
+    graph_objs.
+- genomewideline_width (number; default 1): Width of the genome wide
+    line.
+- highlight (bool; default true): Whether the data points considered
```
mkcor:
```suggestion
- highlight (bool; default True): Whether or not the data points considered
```

dash_bio/component_factory/_volcano.py
```diff
@@ -333,46 +327,45 @@ def figure(
...
-        graph_objs. (Default: "red")
-    - genomewideline_width (number; optional): Width of the genome wide
-      line. (Default: 1)
-    - highlight (bool; optional): Whether the data points considered
-      significant should be highlighted. (Default: True)
-    - highlight_color (string; optional): Color of the data points
+        graph_objs.
+    - genomewideline_width (number; default 1): Width of the genome
+      wide line.
+    - highlight (bool; default true): Whether the data points considered
```
mkcor:
```suggestion
    - highlight (bool; default True): Whether the data points considered
```

dash_bio/component_factory/_volcano.py
```diff
@@ -333,46 +327,45 @@ def figure(
...
         line. Can be in any color format accepted by plotly_js
-        graph_objs. (Default: "red")
-    - genomewideline_width (number; optional): Width of the genome wide
-      line. (Default: 1)
-    - highlight (bool; optional): Whether the data points considered
-      significant should be highlighted. (Default: True)
-    - highlight_color (string; optional): Color of the data points
+        graph_objs.
+    - genomewideline_width (number; default 1): Width of the genome
+      wide line.
```
mkcor: `genome-wide` is more readable

dash_bio/component_factory/_volcano.py
```diff
@@ -333,46 +327,45 @@ def figure(
...
+    - highlight (bool; default true): Whether the data points considered
+      significant should be highlighted.
+    - highlight_color (string; default red): Color of the data points
```
mkcor:
```suggestion
    - highlight_color (string; default 'red'): Color of the data points
```

dash_bio/component_factory/_volcano.py
```diff
@@ -333,46 +327,45 @@ def figure(
...
         highlighted because considered as significant Can be in any color
```
mkcor:
```suggestion
        highlighted because considered as significant. Can be in any color
```

dash_bio/component_factory/_clustergram.py
mkcor: Great! It seems that you fixed the instance line 242 but not this one (line 104).

dash_bio/component_factory/_clustergram.py
```diff
@@ -203,75 +199,72 @@ def Clustergram(
...
-    percentile, would be gray; and the cell(s) in the 100th percentile
-    would be colored black. (Default: [[0.0, 'rgb(255,0,0)'], [0.5,
-    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']])
+    neighboring leaves.
+- color_map (list; default [[0.0, 'rgb(255,0,0)'], [0.5,
+    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']]): The colorscale for the
+    heatmap. Each list element contains two elements; the first
+    element refers to the portion of the maximum data point under
+    which a cell will be colored, and the second element refers to the
+    color. e.g., a colorscale [[0.0, 'white'], [0.5, 'gray'], [1.0,
```
```suggestion
    color, e.g., a colorscale [[0.0, 'white'], [0.5, 'gray'], [1.0,
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -203,75 +199,72 @@ def Clustergram(
...
+    'black']] mean that for all cells with a value below the 50th
```
```suggestion
    'black']] means that for all cells with a value below the 50th
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -203,75 +199,72 @@ def Clustergram(
...
+    percentile of the dataset, the color on the heatmap would be
```
```suggestion
    percentile, the color on the heatmap would be
```
"of the dataset" is confusing; the 50th percentile is the value (not "of the dataset") below which we find half of the observations (given column of the dataset, assuming it's tidy data).

dash_bio/component_factory/_clustergram.py
```diff
@@ -203,75 +199,72 @@ def Clustergram(
...
-    the dataset, the color on the heatmap would be white; all cells
-    with a value in the 50th or higher percentile, excluding the 100th
-    percentile, would be gray; and the cell(s) in the 100th percentile
-    would be colored black. (Default: [[0.0, 'rgb(255,0,0)'], [0.5,
-    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']])
+    neighboring leaves.
+- color_map (list; default [[0.0, 'rgb(255,0,0)'], [0.5,
+    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']]): The colorscale for the
+    heatmap. Each list element contains two elements; the first
+    element refers to the portion of the maximum data point under
```
mkcor: I don't understand the phrase "portion of the maximum data point under which"

dash_bio/component_factory/_clustergram.py
mkcor: > These are meant to mimic the auto-generated docstrings as closely as possible. "True" and "False" are not capitalized in JavaScript, and they don't end up capitalized in the generated Python docstrings either.

I respect the logic, but I think it's better to make docstrings consistent the other way around, the way you suggest:

> I'll commit these suggestions and follow up in the Dash repo that is used to generate Python files so that the React components match!

When writing pure Python components, i.e., writing directly in Python, I don't think we want to mimic their 'inheriting' from JS, just because React components happen to be converted too naively for the time being.

dash_bio/component_factory/_clustergram.py
shammamah: I could instead put "50th-percentile value of the dataset" -- does that make more sense?

dash_bio/component_factory/_clustergram.py
shammamah: These should both be fixed in aa0f5ef771a2a5a26e35c8913664283af06e1742.

dash_bio/component_factory/_clustergram.py
shammamah: Fixed in c377939e8b704e56f452a5ad38e9beeaf53256d7 (I fixed the other docstring, too).

dash_bio/component_factory/_clustergram.py
mkcor: No. First, a percentile is always a value, so it's not necessary to specify "value." It's more standard to say "the 50th percentile." Then, "of the dataset" doesn't make sense.

dash_bio/component_factory/_clustergram.py
```diff
@@ -203,75 +199,72 @@ def Clustergram(
...
+    which a cell will be colored, and the second element refers to the
+    color, e.g., a colorscale [[0.0, 'white'], [0.5, 'gray'], [1.0,
+    'black']] means that for all cells with a value below the 50th
+    percentile of the dataset, the color on the heatmap would be
+    white; all cells with a value in the 50th or higher percentile,
+    excluding the 100th percentile, would be gray; and the cell(s) in
```
```suggestion
    excluding the 100th percentile, would be gray; and the cell(s) at
```
All cells are *in* the 100th percentile (see https://en.wikipedia.org/wiki/Percentile).

dash_bio/component_factory/_clustergram.py
```diff
@@ -203,75 +199,72 @@ def Clustergram(
...
-    would be colored black. (Default: [[0.0, 'rgb(255,0,0)'], [0.5,
-    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']])
+    neighboring leaves.
+- color_map (list; default [[0.0, 'rgb(255,0,0)'], [0.5,
+    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']]): The colorscale for the
+    heatmap. Each list element contains two elements; the first
+    element refers to the portion of the maximum data point under
+    which a cell will be colored, and the second element refers to the
+    color, e.g., a colorscale [[0.0, 'white'], [0.5, 'gray'], [1.0,
+    'black']] means that for all cells with a value below the 50th
```
```suggestion
    'black']] means that for all cells with a value in the 49th
```
would be an alternative

dash_bio/component_factory/_clustergram.py
```diff
@@ -203,75 +199,72 @@ def Clustergram(
...
+    percentile of the dataset, the color on the heatmap would be
```
```suggestion
    percentile, the color on the heatmap would be
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -203,75 +199,72 @@ def Clustergram(
...
+    white; all cells with a value in the 50th or higher percentile,
```
```suggestion
    white; all cells at the 50th or higher percentile,
```
dropping "with a value" like you're doing afterwards

dash_bio/component_factory/_clustergram.py
```diff
@@ -203,75 +199,72 @@ def Clustergram(
...
     will be colored with one shade, and the values that are above this
-    value will be colored with another. (Default: 3)
-- symmetric_value (bool; optional): Whether or not to center the
-    values of the heatmap about zero. (Default: True)
-- log_transform (bool; optional): Whether or not to transforms the
-    data by taking the base-two logarithm of all values in the
-    dataset. (Default: False)
-- display_ratio (list/float; optional): The dendrograms' heights with
+    value will be colored with another.
+- symmetric_value (bool; default true): Whether or not to center the
```
```suggestion
- symmetric_value (bool; default True): Whether or not to center the
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -203,75 +199,72 @@ def Clustergram(
...
+    values of the heatmap about zero.
+- log_transform (bool; default false): Whether or not to transforms
```
```suggestion
- log_transform (bool; default False): Whether or not to transforms
```

dash_bio/component_factory/_volcano.py
```diff
@@ -212,31 +207,30 @@ def __init__(
         numeric.  This does not have to be a p-value. It can be any
         numeric quantity such as peak heights, bayes factors, test
         statistics. If it is not a p-value, make sure to set logp =
-        FALSE. (Default: "P")
+        FALSE.
```
mkcor:
```suggestion
        False.
```

dash_bio/component_factory/_volcano.py
```diff
@@ -333,46 +327,44 @@ def figure(
...
-    - ylabel (string; optional): Label of the y axis. (Default: "-log10(p)")
-    - point_size (number; optional): Size of the points of the Scatter
-      plot. (Default: 5)
+        """Keyword arguments:
+
+    - title (string; default 'Volcano Plot'): Title of the
+        graph.
+    - xlabel (string; optional): Label of the x axis.
+    - ylabel (string; default '-log10(p)'): Label of the y axis.
+    - point_size (number; default 5): Size of the points of the Scatter
```
mkcor:
```suggestion
    - point_size (number; default 5): Size of the points of the scatter
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -203,75 +199,72 @@ def Clustergram(
...
+    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']]): The colorscale for the
+    heatmap. Each list element contains two elements; the first
+    element refers to the portion of the maximum data point under
+    which a cell will be colored, and the second element refers to the
+    color, e.g., a colorscale [[0.0, 'white'], [0.5, 'gray'], [1.0,
+    'black']] means that for all cells with a value in the 49th
+    percentile, the color on the heatmap would be
+    white; all cells at the 50th or higher percentile,
+    excluding the 100th percentile, would be gray; and the cell(s) at
+    the 100th percentile would be colored black.
```
mkcor: Good, we got the percentiles straight in the paragraph above. Now, it would be helpful to keep only one phrasing for all three cases (colours), where this phrasing may 'shrink' as we move forward, but not change completely, e.g.:

* "for values in the 49th percentile, cells would be colored white on the heatmap; for values at the 50th or higher percentile, excluding the 100th percentile, cells would be colored gray; and for values at the 100th percentile, black."
* "for cells with a value in the 49th percentile, the color on the heatmap would be white; for cells with a value at the 50th or higher percentile, excluding the 100th percentile, the color would be gray; and for cells with a value at the 100th percentile, black."

Not going from "cells" to "values," from "color would be" to "would be colored" helps with cognitive load.

dash_bio/component_factory/_clustergram.py
```diff
@@ -67,66 +66,65 @@ def Clustergram(
...
+    neighboring leaves.
+- color_map (list; default [[0.0, 'rgb(255,0,0)'], [0.5,
+    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']]): The colorscale for the
+    heatmap. Each list element contains two elements; the first
+    element refers to the portion of the maximum data point under
+    which a cell will be colored, and the second element refers to the
+    color, e.g., a colorscale [[0.0, 'white'], [0.5, 'gray'], [1.0,
+    'black']] means that for all cells with a value below the 50th
+    percentile of the dataset, the color on the heatmap would be
+    white; all cells with a value in the 50th or higher percentile,
```
```suggestion
    white; all cells with a value at the 50th or higher percentile,
```

dash_bio/component_factory/_clustergram.py
```diff
@@ -67,66 +66,65 @@ def Clustergram(
...
-    would be colored black. (Default: [[0.0, 'rgb(255,0,0)'], [0.5,
-    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']])
+    neighboring leaves.
+- color_map (list; default [[0.0, 'rgb(255,0,0)'], [0.5,
+    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']]): The colorscale for the
+    heatmap. Each list element contains two elements; the first
+    element refers to the portion of the maximum data point under
+    which a cell will be colored, and the second element refers to the
+    color, e.g., a colorscale [[0.0, 'white'], [0.5, 'gray'], [1.0,
+    'black']] means that for all cells with a value below the 50th
```
```suggestion
    'black']] means that for all cells with a value in the 49th
```

shammamah: I'm a bit confused here -- you used "at" and not "in" in a previous suggestion (https://github.com/plotly/dash-bio/pull/379#discussion_r304038202). 

You also mentioned not using the word "value" here: https://github.com/plotly/dash-bio/pull/379#discussion_r298771236

In the interest of consistency, we should make sure everything has the same wording!

mkcor: > I'm a bit confused here -- you used "at" and not "in" in a previous suggestion ([#379 (comment)](https://github.com/plotly/dash-bio/pull/379#discussion_r304038202)).

They mean different things: "in the 49th percentile" means "at the 1st, 2nd, 3rd, ..., 48th, or 49th percentile."

> You also mentioned not using the word "value" here: [#379 (comment)](https://github.com/plotly/dash-bio/pull/379#discussion_r298771236)

Sorry, my wording suggestions at https://github.com/plotly/dash-bio/pull/379#discussion_r304037899 were about the English, the syntax, not the math, the stats anymore.

> In the interest of consistency, we should make sure everything has the same wording!

Absolutely: within each paragraph (those than explain the white/grey/black colours) and between paragraphs (it looks like there are 3 instances of these descriptions).

shammamah: @mkcor Might be faster for you to do it, if that's OK with you!

mkcor: > @mkcor Might be faster for you to do it, if that's OK with you!

@shammamah all good, on it!

dash_bio/component_factory/_manhattan.py
mkcor: Fixed via 1281924045.

mkcor:
It is very time-consuming (and a little frustrating) to review this PR, because many text blocks are featured two, three, or more times: Could/should they not go into some boilerplate, so we can edit/maintain a single source?

mkcor:
I'm aware of #378 so I'm leaving aside those style inconsistencies and, hence, proofreading only the English and Python languages.

dash_bio/component_factory/_clustergram.py
mkcor: I think I got it; rewording in terms of percentile rank (https://en.wikipedia.org/wiki/Percentile_rank).

dash_bio/component_factory/_clustergram.py
```diff
@@ -45,131 +45,120 @@ def Clustergram(
...
-    the dataset, the color on the heatmap would be white; all cells
-    with a value in the 50th or higher percentile, excluding the 100th
-    percentile, would be gray; and the cell(s) in the 100th percentile
-    would be colored black. (Default: [[0.0, 'rgb(255,0,0)'], [0.5,
-    'rgb(0,0,0)'], [1.0, 'rgb(0,255,0)']])
+    column-clustered data.
+- row_dist (string; default 'euclidean'): Distance metric for rows.
+    Passed as argument `metric` to the function specified in `dist_fun`.
+- col_dist (string; default 'euclidean'): Distance metric for columns.
+    Passed as argument `metric` to the function specified in `dist_fun`.
```
mkcor: There's an issue here; either `row_dist` or `col_dist` can be passed as `metric` to the function specified in `dist_fun` (not both).

dash_bio/component_factory/_clustergram.py
```diff
@@ -45,131 +45,120 @@ def Clustergram(
...
+    Top-level elements contain two elements, the first of which refers to
+    the percentile rank, and the second to the applied color. For instance,
+    [[0.0, 'white'], [0.5, 'gray'], [1.0, 'black']] means that cells in the
+    49th percentile would be white; cells at the 50th or higher percentiles,
+    excluding the 100th percentile, would be gray; and the cell(s) at the
+    100th percentile would be black.
 - color_list (dict; optional): The list of colors to use for different
    clusters in the dendrogram that have a root under the threshold for
    each dimension. If there are fewer colors than there are clusters
    along a specific dimension. The keys are: 'row' (for row clusters),
```
mkcor: "If there are ..." ? There's a syntax issue here (in natural language).

dash_bio/component_factory/_clustergram.py
```diff
@@ -45,131 +45,120 @@ def Clustergram(
...
     will be colored with one shade, and the values that are above this
-    value will be colored with another. (Default: 3)
-- symmetric_value (bool; optional): Whether or not to center the
-    values of the heatmap about zero. (Default: True)
-- log_transform (bool; optional): Whether or not to transforms the
-    data by taking the base-two logarithm of all values in the
-    dataset. (Default: False)
-- display_ratio (list/float; optional): The dendrograms' heights with
+    value will be colored with another.
+- symmetric_value (bool; default True): Whether or not to center the
```
mkcor: This isn't a great name for a boolean; you'd rather expect a verb (e.g., `center_values`).

dash_bio/component_factory/_clustergram.py
```diff
@@ -399,9 +276,13 @@ def __init__(
         self._width = width
 
         # convert line width to list if necessary
-        self._line_width = [0, 0]
```
mkcor: I removed this unnecessary line (maybe the intention was to initialize `self._line_width`?).

dash_bio/component_factory/_clustergram.py
shammamah: There is no issue here. The clustering can be performed on both dimensions: https://github.com/plotly/dash-bio/blob/39fec92501c83e4963874125fe4d118d2417d787/dash_bio/component_factory/_clustergram.py#L728

dash_bio/component_factory/_clustergram.py
```diff
@@ -2,7 +2,7 @@
 from random import shuffle
 
 import numpy as np
-import scipy as scp
+import scipy
```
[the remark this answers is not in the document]
shammamah: Let's remove this change from this PR. The scope of this PR is for fixing docstrings.

dash_bio/component_factory/_clustergram.py
shammamah: Fixed in b7a6c49.

dash_bio/component_factory/_clustergram.py
```diff
@@ -201,125 +191,13 @@ def Clustergram(
 
 
 class _Clustergram():
-    """A Dash Bio Clustergram class.
```
shammamah: If we're removing this, we should make reference to the docstring for the "wrapper" `Clustergram()` function above. All of the arguments for that function except `generate_curves_dict`, `return_computed_traces`, and `computed_traces` are eventually passed here.

mkcor: Right, I double-checked every "keyword argument;" except for `generate_curves_dict`, `return_computed_traces`, and `computed_traces` indeed, they are parameters of `_Clustergram()`'s **methods**, so they could/should be documented in each corresponding method's docstring. Would you like me to do this in this PR, or in a different PR?

shammamah: Let's do this in a different PR.

dash_bio/component_factory/_clustergram.py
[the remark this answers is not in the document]
mkcor: Done via 707c8d2.

dash_bio/component_factory/_manhattan.py
```diff
@@ -1,54 +1,17 @@
 from __future__ import absolute_import
 
+import numpy as np
 import pandas as pd
 from pandas.api.types import is_numeric_dtype
-import numpy as np
 
 import plotly.graph_objs as go
 
+from ._volcano import _get_hover_text
```
shammamah: I'm a bit conflicted about leaving this in here, since the two plots don't really have anything to do with each other besides this shared functionality. If they're importing it from somewhere, it should come from a third file (in my opinion).

dash_bio/component_factory/_manhattan.py
```diff
@@ -219,37 +177,35 @@ def __init__(
         - chrm (string; optional): A string denoting the column name for the
```
shammamah:
```suggestion
        - chrm (string; default 'CHR'): A string denoting the column name for the
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -219,37 +177,35 @@ def __init__(
...
         chromosome.  This column must be float or integer.  Minimum number
         of chromosomes required is 1. If you have X, Y, or MT chromosomes,
-        be sure to renumber these 23, 24, 25, etc. (Default: "CHR")
+        be sure to renumber these 23, 24, 25, etc.
         - bp (string; optional): A string denoting the column name for the
```
shammamah:
```suggestion
        - bp (string; default 'BP'): A string denoting the column name for the
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -219,37 +177,35 @@ def __init__(
...
-        chromosomal position. (Default: "BP")
+        chromosomal position.
         - p (string; optional): A string denoting the column name for the
```
shammamah:
```suggestion
        - p (string; default 'P'): A string denoting the column name for the
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -219,37 +177,35 @@ def __init__(
...
         float quantity to be plotted on the y-axis. This column must be
         numeric. This does not have to be a p-value. It can be any
         numeric quantity such as peak heights, bayes factors, test
         statistics. If it is not a p-value, make sure to set logp = FALSE.
-        (Default: "P")
         - snp (string; optional): A string denoting the column name for the
```
shammamah:
```suggestion
        - snp (string; default 'SNP'): A string denoting the column name for the
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -219,37 +177,35 @@ def __init__(
...
         SNP names (e.g. rs number). More generally, this column could be
         anything that identifies each point being plotted. For example, in
         an Epigenomewide association study (EWAS) this could be the probe
         name or cg number. This column should be a character. This
         argument is optional, however it is necessary to specify if you
         want to highlight points on the plot using the highlight argument
-        in the figure method. (Default: "SNP")
+        in the figure method.
         - gene (string; optional): A string denoting the column name for the
```
shammamah:
```suggestion
        - gene (string; default 'GENE'): A string denoting the column name for the
```

dash_bio/component_factory/_manhattan.py
```diff
@@ -219,37 +177,35 @@ def __init__(
...
-        generally this could be any annotation information that you want
-        to include in the plot. (Default: "GENE")
+        generally, it could be any annotation information that you want
+        to include in the plot.
         - annotation (string; optional): A string denoting the column name for
         an annotation. This column could be a string or a float.  This
         could be any annotation information that you want to include in
         the plot (e.g. zscore, effect size, minor allele frequency).
-        (Default: None)
         - logp (bool; optional): If True, the -log10 of the p-value is
```
shammamah:
```suggestion
        - logp (bool; default True): If True, the -log10 of the p-value is
```

dash_bio/component_factory/_clustergram.py
mkcor: I mean, the understanding is ambiguous for the end user (when reading only the docstring, as I experienced it). Now that you pointed to the relevant code block (thanks :)), I can understand clearly.

Looking at the code, we can see that a given metric may be used for the row distance (speaking elliptically):
https://github.com/plotly/dash-bio/blob/39fec92501c83e4963874125fe4d118d2417d787/dash_bio/component_factory/_clustergram.py#L734
while another metric may be used for the column distance:
https://github.com/plotly/dash-bio/blob/39fec92501c83e4963874125fe4d118d2417d787/dash_bio/component_factory/_clustergram.py#L730

But the distance function is the same, it's passed once and for all to `self._dist_fun`:
https://github.com/plotly/dash-bio/blob/39fec92501c83e4963874125fe4d118d2417d787/dash_bio/component_factory/_clustergram.py#L242

(Following the logic of a row distance vs. a column distance, we could imagine a 'row distance function' vs. a 'column distance function' -- my confusion rose from that flexibility in the distance choice but not in the distance function choice.)

I think that the following addition would clear up any confusion:
```
- row_dist (string; default 'euclidean'): Distance metric for rows.
    Passed as argument `metric` to the function specified in `dist_fun`
    when called for clustering along rows.
- col_dist (string; default 'euclidean'): Distance metric for columns.
    Passed as argument `metric` to the function specified in `dist_fun`
    when called for clustering along columns.
```

dash_bio/component_factory/_manhattan.py
mkcor: How about moving this helper function to the `dash-bio-utils` package?

dash_bio/component_factory/_clustergram.py
shammamah: That looks good to me! :)

dash_bio/component_factory/_clustergram.py
mkcor: Ok.

dash_bio/component_factory/_manhattan.py
shammamah: We shouldn't do that, unless we're willing to include that package as a dependency for this package (which is the opposite of what we wanted). It's necessary for the *components* to work; the `dash-bio-utils` package is only necessary for the particular set of demo apps that we've created in this repo.

mkcor: Right, right, that's not the spirit. How about creating a `utils.py` module in here (`dash_bio/component_factory/`) then?

shammamah: I think that makes the most sense!

mkcor: Anyway, I would argue that Manhattan plots and Volcano plots do have something to do with each other: they can both make use of a hover text utility displaying 'SNP' and 'GENE' stuff so, apparently, evidently, clearly, they are used to represent a certain kind of datasets, i.e., those datasets with 'SNP' and 'GENE' data (which is already a lot to have in common).

The docstring starts by stating that it applies to both Manhattan and Volcano plots (already in the original submission, this is not something that I changed).

dash_bio/component_factory/_clustergram.py
mkcor: Done in 077c84a698b5a :)

dash_bio/component_factory/_manhattan.py
mkcor: Done via 6675dc30bc69441e165f00f5a6468e7fd7431d3a.

dash_bio/component_factory/_clustergram.py
mkcor: addressing in PR #401

PR #381, opened 2019-06-27
Mol2d proptypes fix

shammamah:
## About 

* Changed dependency for molecule-2d-for-react to the Plotly fork, which uses external `prop-types`.

PR #314, opened 2019-04-10
Updated dependencies and PropTypes fix

shammamah:
Closes #313 

## About 
- [ ] This is a new component 
- [ ] I am adding a feature to an existing component, or improving an existing feature
- [x] I am closing an issue 

## Description of changes 
* I transferred ownership of all of the forked repos that were under my account to Plotly, and changed the dependency locations to reflect that. 
* In the `molecule-3d-for-react` fork I made, I addressed https://github.com/Autodesk/molecule-3d-for-react/issues/36 and it no longer gives us the error I was getting. 

## Before merging
- [x] I have gone through the [code review checklist](https://github.com/plotly/dash-component-boilerplate/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/review_checklist.md)
- [x] I have completed all of the [steps to take before merging](https://github.com/plotly/dash-bio/blob/master/.github/before_merging.md)
- [x] I know how to [deploy to DDS](https://github.com/plotly/dash-bio/blob/master/.github/deploying_to_dds.md)

PR #382, opened 2019-06-27
Update dashBio package for compatibility with dash 0.1.0

rpkyle:
## About
- I am updating the R package which provides `dash-bio` support in Dash for R.

## Description of changes
- Implements support for externalized `prop-types` #314 
- Implements Mol2d propTypes fix #381  
- Amends existing documentation, adds updated package info file for R

## Before merging


DESCRIPTION
```diff
@@ -1,15 +1,15 @@
 Package: dashBio
 Title: Dash components for bioinformatics
-Version: 0.0.11
+Version: 0.1.2
 Authors @R: as.person(c(The Plotly Team <dashbio@plot.ly>))
 Description: Dash components for bioinformatics
 Depends: R (>= 3.0.2), manhattanly
 Imports: dash
-Suggests:
+Suggests:
```
shammamah: Should this whitespace be here?

man/dashbioCircos.Rd
```diff
@@ -39,16 +39,75 @@ selectEvent={
...
     },}
 
-\item{layout}{List. The overall layout of the Circos graph, provided
+\item{layout}{List of lists containing elements 'len', 'color', 'label', 'id'.
+those elements have the following types:
+  - len (numeric; required): the length of the block.
+  - color (character; required): the color of the block.
+  - label (character; required): the labels of the block.
+  - id (character; required): the id of the block, where it will recieve
+data from the specified "track" id.s. The overall layout of the Circos graph, provided
```
shammamah: It might be better to have everything after/including "The overall layout..."  before this list -- it kind of looks like it's part of the last element in the list, but it isn't. The extraneous `s.` should be covered by my changes in https://github.com/plotly/dash/pull/782, but those haven't been released yet.

DESCRIPTION
rpkyle: Can fix that!

man/dashbioCircos.Rd
rpkyle: Yeah, this is a tricky one (but a very valid point). It's because the generator detects that `layout` has `PropTypes.exact`, and so it handles the nested typing. Because it's a typing argument, it's at the head of the string (as are all types).

I'm not entirely sure how we would break the typing string in half, and insert the _The overall layout of the Circos graph, provided as a list of dictionaries._ after _..., 'id'._ but I can look into it for the next release. I'll open an issue in the `dash` repo so I don't lose track of this.

rpkyle: Thanks for the feedback; noted in https://github.com/plotly/dash/issues/799.

.Rbuildignore
```diff
@@ -17,12 +22,13 @@ demo/.*\.css
 
 # ignore python files/folders
 setup.py
-index.py
-config.py
+usage.py
+setup.py
```
mkcor: Looks like a regression... I remember reviewing this before, but I can't break the commits down because it looks like they were rebased into 28c375bea4306750.

.Rbuildignore
```diff
@@ -17,12 +22,13 @@ demo/.*\.css
...
 requirements.txt
 MANIFEST.in
 CHANGELOG.md
-tests/
-.pylintrc
+test/
+# CRAN has weird LICENSE requirements
+LICENSE.txt
```
mkcor: Same here, I remember reviewing this and @shammamah had fixed it accordingly.

.Rbuildignore
rpkyle: Thanks for noting this. Fixed in d7bd0d46f052ffe8a50697947ef7bddc7194854f.

.Rbuildignore
rpkyle: Fixed in d7bd0d46f052ffe8a50697947ef7bddc7194854f

DESCRIPTION
rpkyle: Fixed in 26f04011389f45d1c94438ff7830fa194454e93c.

PR #386, opened 2019-07-03
Sequence viewer issue

### LINKED ISSUE #385: Sequence Viewer app doesn't save selection/coverage data due to dcc.Loading component
To fix: Remove the wrapping `dcc.Loading` from the `SequenceViewer` component in `app_sequence_viewer.py`.

celinehuang:
Closes: #385


tests/dashbio_demos/app_sequence_viewer.py
```diff
@@ -69,8 +69,8 @@ def description():
 
 def layout():
 
-    return html.Div(id='seq-view-body', className='app-body', children=[
-        dcc.Loading(className='dashbio-loading', children=html.Div(
+    return html.Div(id='seq-view-body', className='app-body dashbio-loading', children=[
```
shammamah:
```suggestion
    return html.Div(id='seq-view-body', className='app-body', children=[
```

We don't need the loading state class since we're getting rid of the `dcc.Loading` component!

PR #411, opened 2019-08-01
Dashbio tests refactoring

### LINKED ISSUE #103: Write tests for all components/apps

### LINKED ISSUE #416: Molecule2dViewer cannot write to selectedAtomIds except upon initial render
**Describe the bug**
Defining the value of the `selectedAtomIds` prop through a callback does not cause the corresponding atoms to be highlighted in the graph. 

**To Reproduce**
- create an app with the `Molecule2dViewer` component and don't define the value of `selectedAtomIds`
- write a callback that modifies `selectedAtomIds` according to the value of some input (e.g., a dropdown) 

**Expected behavior**
- the atoms that correspond to the ids selected in the dropdown should be highlighted

shammamah:
Closes #103 (finally!)

Closes #416.

## About
- [ ] This is a new component
- [x] I am closing an issue

## Description of changes
* Add new Python test suite that uses `dash[testing]` for integration tests for all components
* Add initial boilerplate code for unit tests (to be completed in a later PR; right now, the tests only check whether or not the JS elements render) 
* Add Percy integration
* Fix some bugs in components (see CHANGELOG updates in d6a8bf5)

### Still remaining to complete (in future PRs): 
* Add more unit tests for JS components
* Capture canvas elements (they are not included in the DOM, so they are not captured by Percy)

## Before merging


tests/requirements.txt
```diff
@@ -7,5 +7,5 @@ flake8==3.6.0
 ipdb==0.11
 percy==2.0.1
 pylint==2.3.0
-pytest-dash==2.0.0
+pytest==4.6.4
```
byronz: Please refer to https://github.com/plotly/dash/blob/dev/requires-testing.txt to remove the redundants

dash_bio/Circos.py
```diff
@@ -30,14 +30,74 @@ class Circos(Component):
...
+
+The key "name" is required for this dictionary,
+where the input for "name" points to some list of
+dictionaries color values.
+
+Ex: "color": {"name": "some key that refers to color in a data set"}. color has the following type: a value equal to: PropTypes.string, PropTypes.shape({
+    name: PropTypes.string.isRequired,
+})"""
     @_explicitize_args
     def __init__(self, enableDownloadSVG=Component.UNDEFINED, enableZoomPan=Component.UNDEFINED, id=Component.UNDEFINED, style=Component.UNDEFINED, eventDatum=Component.UNDEFINED, selectEvent=Component.UNDEFINED, layout=Component.REQUIRED, config=Component.UNDEFINED, size=Component.UNDEFINED, tracks=Component.UNDEFINED, **kwargs):
```
mkcor: Could we regenerate this file in order to introduce line breaks here?

dash_bio/Ideogram.py
```diff
@@ -130,21 +135,21 @@ class Ideogram(Component):
...
+- rangeSet (list of dicts; optional): List of objects describing segments of recombination among chromosomes in a chromosome set.
+- rotatable (boolean; default True): Whether chromosomes are rotatable on click.
 - rotated (boolean; optional): Dash callback that returns true if rotated, and false if not.
 - sex (a value equal to: 'male', 'female'; optional): Useful for omitting chromosome Y in female animals.
 Currently only supported for organisms that use XY sex-determination.
-- showChromosomeLabels (boolean; optional): Whether to show chromosome labels, e.g., 1, 2, 3, X, Y.
-- showBandLabels (boolean; optional): Whether to show cytogenetic band labels, e.g., 1q21
-- showFullyBanded (boolean; optional): Whether to show fully banded chromosomes for genomes that have sufficient data. Useful for
+- showChromosomeLabels (boolean; default True): Whether to show chromosome labels, e.g., 1, 2, 3, X, Y.
+- showBandLabels (boolean; default False): Whether to show cytogenetic band labels, e.g., 1q21
```
mkcor:
```suggestion
- showBandLabels (boolean; default False): Whether to show cytogenetic band labels, e.g., 1q21.
```

shammamah: Could you make these suggestions in the source file (`src/lib/components/Ideogram.react.js`) instead? The Python files are generated from the JS ones.

dash_bio/Circos.py
shammamah: We can try modifying the JS file. If that doesn't work, it would have to go somewhere in https://github.com/plotly/dash/blob/dev/dash/development/_py_components_generation.py.

dash_bio/Ideogram.py
```diff
@@ -87,15 +92,15 @@ class Ideogram(Component):
...
 You will need to set up your own database to grab data from a custom database.
-- organism (string | number; optional): Organism(s) to show chromosomes for. Supply organism's name as a string (e.g., "human") or
+- organism (string | number; default 'human'): Organism(s) to show chromosomes for. Supply organism's name as a string (e.g., "human") or
 organism's NCBI Taxonomy ID (taxid, e.g., 9606) to display chromosomes from a single
 organism, or an array of organisms' names or taxids to display chromosomes from multiple
 species.
 - localOrganism (dict; optional): Provide local JSON organism into this prop from a local user JSON file.
 DataDir must not be initialized.
-- homology (optional): Used to compare two chromosomes with each other.
+- homology (dict; optional): Used to compare two chromosomes with each other.
```
mkcor:
```suggestion
- homology (dict; optional): Used to compare two chromosomes from different organisms.
```
Repeating the definition of homology... Alternatively, we may end the description after "chromosomes."

dash_bio/Molecule3dViewer.py
```diff
@@ -13,12 +13,15 @@ class Molecule3dViewer(Component):
...
 - id (string; optional): The ID used to identify this component in callbacks
-- selectionType (a value equal to: 'atom', 'residue', 'chain'; optional): The selection type - may be atom, residue or chain
-- backgroundColor (string; optional): Property to change the background color of the molecule viewer
-- backgroundOpacity (number; optional): Property to change the backgroun opacity - ranges from 0 to 1
-- styles (list; optional): Property that can be used to change the representation of
-the molecule. Options include sticks, cartoon and sphere
-- modelData (optional): The data that will be used to display the molecule in 3D
+- selectionType (a value equal to: 'atom', 'residue', 'chain'; default 'atom'): The selection type - may be atom, residue or chain
+- backgroundColor (string; default '#FFFFFF'): Property to change the background color of the molecule viewer
+- backgroundOpacity (number; default 0): Property to change the backgroun opacity - ranges from 0 to 1
```
mkcor:
```suggestion
- backgroundOpacity (number; default 0): Property to change the background opacity - ranges from 0 to 1
```

dash_bio/Molecule3dViewer.py
```diff
@@ -13,12 +13,15 @@ class Molecule3dViewer(Component):
...
+- styles (dict; optional): Property that can be used to change the representation of
+the molecule. Options include sticks, cartoon and sphere. styles has the following type: list of dicts containing keys 'color', 'visualization_type'.
+Those keys have the following types:
```
mkcor:
```suggestion
These keys have the following types:
```

shammamah: This also needs to go in https://github.com/plotly/dash/blob/fbb2afcc61395abd5a3d63408f4dfddc6305bebd/dash/development/_py_components_generation.py#L479.

tests/integration/test_manhattan_plot.py
```diff
@@ -0,0 +1,178 @@
...
+from common_features import nested_component_layout, \
+    nested_component_app_callback
+
+_data = pandas.read_csv(
+    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/' +
+    'manhattan_data.csv'
+)
+
+
+def test_dbmp001_genomewideline_value(dash_duo):
```
byronz: this is a comment for all the tests here in general, I feel most of the tests share the same pattern, or the same common function, this is a sign for `pytest.mark.parametrize()`

dash_bio/Ideogram.py
mkcor: Right, I thought of it afterwards.

dash_bio/Molecule3dViewer.py
mkcor: Oh, that's where it's coming from. Thanks for pointing me to the source.

dash_bio/Ideogram.py
mkcor: Actually, `src/lib/components/Ideogram.react.js` isn't part of the changesets included in this PR... How come?

CHANGELOG.md
```diff
@@ -1,5 +1,40 @@
...
+* Added re-rendering for Molecule2dViewer and Molecule3dViewer upon
+  the `selectedAtomIds` prop being updated. (This allows for the user
+  to write to the `selectedAtomIds` prop through a Dash component.)
+* Added re-rendering for SequenceViewer upon the `coverageClicked`
+  prop being updated.
+
+## [0.1.4] - 2019-08-02
+
+### Fixed
+* Fixed incompatibility with Dash 1.0.0.
```
mkcor: Hmm, I would put this under "Changed" and call it "Migrated to React 16 and Dash 1" :)

CHANGELOG.md
```diff
@@ -1,5 +1,40 @@
...
+  within the `links` property to also include `PropTypes.shape`.
+* Fixed Molecule3dViewer error in which there was no re-rendering upon
+  a change in the `selectionType` prop.
+* Fixed SequenceViewer error in which the component initially rendered
+  with coverage that had no defined `onClick` behavior.
+
+### Added
+
+* Added re-rendering for Molecule2dViewer and Molecule3dViewer upon
+  the `selectedAtomIds` prop being updated. (This allows for the user
```
```suggestion
  the `selectedAtomIds` prop being updated. (This lets the user
```

CHANGELOG.md
```diff
@@ -1,5 +1,40 @@
...
+  to write to the `selectedAtomIds` prop through a Dash component.)
```
```suggestion
  write to the `selectedAtomIds` prop through a Dash component.)
```
("allows for" expects a noun)

tests/unit/unit_test_data_setup.py
```diff
@@ -0,0 +1,26 @@
+import json
+
+from dash_bio_utils import pdb_parser, styles_parser, \
+    protein_reader, xyz_reader
```
mkcor: We might as well import in alphabetical order.

.circleci/config.yml
```diff
@@ -46,13 +46,24 @@ jobs:
...
+                    npm run test
 
     "python-3.6": &test-template
         docker:
-            - image: circleci/python:3.6-stretch-browsers
+            - image: circleci/python:3.6-stretch-node-browsers
 
         environment:
-            PERCY_ENABLED: False
+            PERCY_ENABLED: True
```
mkcor: Wow! :)

.circleci/config.yml
```diff
@@ -92,24 +104,27 @@ jobs:
...
+                name: Run integration tests
                 command: |
                     . venv/bin/activate
                     pip install --upgrade pip
                     pip install -r tests/requirements.txt
-                    pytest --webdriver Chrome tests --ignore tests/test_skeleton.py
+                    npm i
+                    npm run build:all-dev
+                    pip install -e .
+                    pytest --webdriver Chrome tests/integration
```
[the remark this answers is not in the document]
mkcor: I can't run the tests locally (`argparse.ArgumentError: argument --webdriver: conflicting option string: --webdriver`), I'll troubleshoot later...

dash_bio/Ideogram.py
shammamah: Because that file didn't change at all. The reason the Python files changed is due to the updated docstring formatting introduced in https://github.com/plotly/dash/pull/782.

CHANGELOG.md
shammamah: I think it makes more sense to put it under "fixed", since this library can't really be used without Dash and it *is* an error for it to be out-of-date.

dash_bio/Circos.py
mkcor: Absolutely! I'll resolve this conversation and address the issue separately.

mkcor: Issue lives at https://github.com/plotly/dash/issues/931.

CHANGELOG.md
Marc-Andre-Rivet: Would make this `Changed`. Moving from `dash==0.40` to `dash>=1.0.0` is not a bug fix but updating one of the library's core dependency. As such,React 16 can be mentioned but is just a side-effect of updating Dash compatibility and I would personally skip it.

src/lib/components/AlignmentChart.react.js
```diff
@@ -35,6 +35,11 @@ export default class AlignmentChart extends Component {
     render() {
         const {id, eventDatum} = this.props;
 
+        // TODO handle this in react-alignment-viewer instead
+        if (this.props.colorscale === null) {
+            this.props.colorscale = 'clustal2';
+        }
```
Marc-Andre-Rivet: Playing with props this way makes me a bit queazy. Could be done alongside the omit

```
                {...{
                    ...omit(
                        ['fireEvent', 'dashEvent', 'setProps'],
                        this.props
                    )},
                    colorscale: this.props.colorscale === null ? 'clustal2': this.props.colorscale === null
                }}
```
or some variation

src/lib/components/Circos.react.js
```diff
@@ -173,32 +173,34 @@ export default class Circos extends Component {
...
-                configApply = config;
+                if (typeof config !== 'undefined') {
+                    configApply = config;
 
-                // Set Event Handling
-                if (setProps) {
-                    configApply.events = this.setEvent(setProps, index);
-                }
+                    // Set Event Handling
+                    if (setProps) {
```
Marc-Andre-Rivet: `setProps` check isn't necessary anymore

src/lib/components/Circos.react.js
```diff
@@ -173,32 +173,34 @@ export default class Circos extends Component {
...
+                const {id, data, config, type} = track;
 
-            // Since config is const, can't manipulate and throws error
-            let configApply;
+                // Since config is const, can't manipulate and throws error
+                let configApply;
 
-            if (typeof config !== 'undefined') {
-                configApply = config;
+                if (typeof config !== 'undefined') {
```
Marc-Andre-Rivet: is es-lint preventing you from using ```undefined```?

src/lib/components/Molecule2dViewer.react.js
```diff
@@ -21,7 +21,23 @@ export default class Molecule2dViewer extends Component {
...
+                    nextProps.selectedAtomIds.length) ||
+            (this.props.selectedAtomIds &&
+                nextProps.selectedAtomIds &&
+                (this.props.selectedAtomIds.some(
+                    atomId => !(atomId in nextProps.selectedAtomIds)
+                ) ||
+                    nextProps.selectedAtomIds.some(
+                        atomId => !(atomId in this.props.selectedAtomIds)
+                    )))
+        ) {
```
Marc-Andre-Rivet: No obligation here but would suggest flattening this for readability.
Same comment applies to other `shouldComponentUpdate`.

```
shouldComponentUpdate(nextProps) {
    if (this.props.modelData !== nextProps.modelData) {
        return true;
    }

    // if one is defined but not the other (XOR)
    if (this.props.selectedAtomIds ? !nextProps.selectedAtomIds : nextProps.selectedAtomIds) {
        return true;
    }

    // etc.
}
```

src/lib/components/SequenceViewer.react.js
```diff
@@ -147,8 +148,18 @@ export default class SequenceViewer extends Component {
             legend: this.props.legend,
         };
 
+        const coverage = this.props.coverage;
+
+        for (let i = 0; i < coverage.length; i++) {
+            coverage[i].onclick = this.getOnClick(i);
+        }
```
Marc-Andre-Rivet: Is there a good reason for doing this directly on a prop vs. mapping onto new objects?

src/lib/components/AlignmentChart.react.js
shammamah: @Marc-Andre-Rivet Would it be sufficient to just treat the `colorscale` prop as a boolean? e.g., 

`colorscale={this.props.colorscale ? this.props.colorscale : 'clustal2'}`

src/lib/components/Circos.react.js
shammamah: Yep: 

```
  183:32  error  Unexpected use of undefined  no-undefined
```

Is it better to disable that error on this line?

shammamah: Yep: 

```
  183:32  error  Unexpected use of undefined  no-undefined
```

Is it better to disable that error on this line?

src/lib/components/Molecule2dViewer.react.js
shammamah: Opened an issue here: #419.

src/lib/components/Circos.react.js
Marc-Andre-Rivet: This eslint rule exists to prevent doing checks against `undefined`. This is only relevant if targeting browsers supporting ES3 (which we are not, and never will do) -- all moderately modern browsers fully support the ES5 standard which guarantees that `undefined` is read-only.

In ES3 you could do something like `undefined = 'potato'` and all subsequent checks against it would have been, well, potato..

You can remove this rule as it is useless. We've done the same, for the same reasons not long ago in the renderer.

src/lib/components/Circos.react.js
Marc-Andre-Rivet: `Ideogram.react.js` also does a `setProps` check. Either for now or as a subsequent clean up PR.

src/lib/components/AlignmentChart.react.js
Marc-Andre-Rivet: This is fine if what you wanted to do was check truthiness not `null` specifically. This behaves differently with `undefined` and `0`.

src/lib/components/SequenceViewer.react.js
shammamah: Fixed in 79f3899!

src/lib/components/Circos.react.js
shammamah: Fixed in e17d9b0.

src/lib/components/AlignmentChart.react.js
shammamah: The behaviour I want is for the default to be `'clustal2'`, which means that it should revert to that whenever a colorscale isn't specified (which applies to `''`, `undefined`, and `null`). Please correct me if I'm wrong!

src/lib/components/Circos.react.js
shammamah: Fixed in 03131de!

src/lib/components/SequenceViewer.react.js
```diff
@@ -147,8 +126,18 @@ export default class SequenceViewer extends Component {
             legend: this.props.legend,
         };
 
+        const coverageWithClicks = this.props.coverage.map((entry, i) => {
+            return Object.assign(this.props.coverage[i], {
```
Marc-Andre-Rivet:
```suggestion
            return Object.assign({}, entry, {
```

Marc-Andre-Rivet: The first param is the target, unlike `R.merge` that returns a new object.

dash_bio/Ideogram.py
mkcor: Oh, ok! I see that a lot is happening in this PR... not just the migration of the test suite, which is already something substantial... Anyway, I'll remove these suggestions, they will go into a separate PR.

tests/requirements.txt
mkcor: And do we still want `--extra-index-url https://test.pypi.org/simple/` (first line of this file)?

tests/integration/test_manhattan_plot.py
shammamah: @byronz It appears to me as though the `pytest.mark.parametrize()` feature should be used when testing the same parameter with different values -- the tests in this PR test multiple parameters, but each with only one value. (This is obviously something that I'll be fixing in the future, and I think it would be appropriate to use this suggestion at that time. Please correct me if I'm wrong!)

byronz: you can do it with multiple *parameters*, but that also means the result permutation is significantly big (it's a like a nested loop).  I just pointed out one possible alternative, but it's totally up to you to structure the tests :)

PR #418, opened 2019-09-18
Add first draft of FornaContainer component.

### LINKED ISSUE #417: Secondary analysis component
https://github.com/ViennaRNA/forna

>Dash Bio has visualizations for primary analysis (electrical signal to nucleic acid sequence readings) and tertiary analysis (protein creation from RNA), but not secondary (base‐pairing of complementary nucleotides).

This would involve going back into the d3 force-directed graph and wrapping that in a React component for Dash Bio.

shammamah:
Closes #417.

## About
- [x] This is a new component
- [ ] I am closing an issue

## Before merging


src/lib/components/FornaContainer.react.js
```diff
@@ -0,0 +1,206 @@
...
+            (sequences && !nextProps.sequences) ||
+            (!sequences && nextProps.sequences) ||
+            sequences.length !== nextProps.sequences.length ||
+            sequences.some(
+                (seq, i) =>
+                    sequences[i].sequence !== nextProps.sequences[i].sequence ||
+                    sequences[i].structure !==
+                        nextProps.sequences[i].structure ||
+                    (sequences[i].options && !nextProps.sequences[i].options) ||
+                    (!sequences[i].options && nextProps.sequences[i].options) ||
```
alexcjohnson: A pattern I like for `(a && !b) || (!a && b)` is `(Boolean(a) !== Boolean(b))` (or `(!!a !== !!b)` but I some of our linters disapprove of `!!`)

Also https://ramdajs.com/docs/#zip would likely clean this section up a lot

src/lib/components/FornaContainer.react.js
```diff
@@ -0,0 +1,206 @@
...
+     * The width (in px) of the container in which the molecules will
+     * be displayed.
+     */
+    width: PropTypes.number,
+
+    /**
+     * The molecules that will be displayed.
+     */
+    sequences: PropTypes.arrayOf(
+        PropTypes.shape({
```
alexcjohnson: `PropTypes.exact` instead of `shape`?

shammamah: The user doesn't have to specify a value for `options`, so `PropTypes.shape` applies here (please correct me if I'm wrong)!

alexcjohnson: yeah the naming is tricky but I think `exact` is what you want (in fact we almost never want `shape`). The only difference, in my understanding, is whether extra, non-enumerated keys are allowed - `shape` would not give you a warning/error if you include other things in the object, but `exact` will. `.isRequired` is still honored either way - so you can send an object to `exact` that's missing some optional keys.

From https://www.npmjs.com/package/prop-types:
```js
// An object taking on a particular shape
  optionalObjectWithShape: PropTypes.shape({
    optionalProperty: PropTypes.string,
    requiredProperty: PropTypes.number.isRequired
  }),
 
  // An object with warnings on extra properties
  optionalObjectWithStrictShape: PropTypes.exact({
    optionalProperty: PropTypes.string,
    requiredProperty: PropTypes.number.isRequired
  }),
```

src/lib/components/FornaContainer.react.js
```diff
@@ -0,0 +1,274 @@
...
+    }
+}
+
+FornaContainer.propTypes = {
+    /**
+     * The ID of this component, used to identify dash components in
+     * callbacks. The ID needs to be unique across all of the
+     * components in an app.
+     */
+    id: PropTypes.string.isRequired,
```
alexcjohnson: If I'm reading this right, `FornaContainer` can accept either a selector string or a DOM node (it's just getting [passed to `d3.select`](https://github.com/ViennaRNA/fornac/blob/master/app/scripts/fornac.js#L802)). If you use a `ref` and pass that instead of `'#' + id`, then id becomes optional, making it easier to use and more robust.

src/lib/components/FornaContainer.react.js
```diff
@@ -0,0 +1,274 @@
...
+                    Boolean(seq[0].options) &&
+                    Boolean(seq[1].options) &&
+                    Object.keys(seq[0].options).some(
+                        _option =>
+                            seq[0].options._option !== seq[1].options._option
+                    )
+            )
+        ) {
+            return true;
+        }
```
alexcjohnson: seems like this could be more efficient combining all the loops into one. Also doesn't the whole `symmetricDifference` then looping through options - in fact even the `Boolean(...) !== Boolean(...)` amount to "are the two objects equal"? Which is what `R.equals` is made for:

```js
sequence_compare.some(
    ([prev, next]) =>
        prev.sequence !== next.sequence ||
        prev.structure !== next.structure ||
        !R.equals(prev.options, next.options)
)
```

or in fact that's everything in the `sequences` prop, right? You just want to know if there's any change anywhere in that prop... So can't this WHOLE thing - all the way up to `if (Boolean(sequences) !== Boolean(nextProps.sequences))` - be replaced by
```js
if (!R.equals(sequences, nextProps.sequences)) {
    return true;
}
```
? Apologies for not noticing that earlier, before you went through a bunch of iterations here...

src/lib/components/FornaContainer.react.js
```diff
@@ -0,0 +1,274 @@
...
+            this._fornaContainer.setOutlineColor(nextProps.nodeFillColor);
+        }
+
+        this._fornaContainer.changeColorScheme(colorScheme);
+
+        return false;
+    }
+
+    render() {
+        this.renderNewSequences();
```
alexcjohnson: Seems a little weird to put this in `render` - so it'll do nothing on initial render, because `if (this._fornaContainer)` will fail, but then on updates it happens *first*, then we depend on react keeping the same element in place as it rerenders.

Might be more natural to put this in `componentDidUpdate`, which would be more analogous to `componentDidMount` where the sequence rendering is handled on first render.

tests/dashbio_demos/app_circos.py
```diff
@@ -1130,8 +1128,8 @@ def layout():
                                 'marginTop': '5px',
                                 'marginBottom': '10px',
                             },
-                            n_fixed_rows=1,
-                            n_fixed_columns=1
+                            fixed_rows=1,
+                            fixed_columns=1
```
alexcjohnson: closer... in addition to the name change, these props are now objects `{headers: bool, data: int}`

tests/dashbio_demos/app_forna_container.py
```diff
@@ -0,0 +1,585 @@
...
+                                html.H4(className='what-is', children='What is FornaContainer?'),
+                                html.P('FornaContainer is a force-directed graph that is '
+                                       'used to represent the secondary structure of nucleic '
+                                       'acids (i.e., DNA and RNA).'),
+                                html.P('In the "Add New" tab, you can enter a sequence '
+                                       'by specifying the nucleotide sequence and the '
+                                       'dot-bracket representation of the secondary '
+                                       'structure.'),
+                                html.P('In the "Sequences" tab, you can select which '
+                                       'sequences will be displayed, as well as obtian '
```
alexcjohnson:
```suggestion
                                       'sequences will be displayed, as well as obtain '
```

could this section be simplified with `dcc.Markdown`?

tests/dashbio_demos/app_forna_container.py
```diff
@@ -0,0 +1,585 @@
...
+                )
+            ]),
+
+            dcc.Store(id='forna-sequences', data=initial_sequences),
+            dcc.Store(id='forna-custom-colors')
+        ]
+    )
+
+
+def callbacks(app):  # pylint: disable=redefined-outer-name
```
alexcjohnson: is this `pylint` directive necessary?

src/lib/components/FornaContainer.react.js
shammamah: Yes, that makes sense! I didn't know about `R.equals`, looks like I can use it in a lot of other components to (like `SequenceViewer`).

tests/dashbio_demos/app_forna_container.py
shammamah: Yes, it's in all of the apps. It's because we have the option to either run them as standalone or as "sub-apps" of the app defined in `index.py`: 

https://github.com/plotly/dash-bio/blob/729402c5e20dcd0425bf5bbe52dedad99a93721f/tests/dashbio_demos/app_clustergram.py#L901
https://github.com/plotly/dash-bio/blob/729402c5e20dcd0425bf5bbe52dedad99a93721f/index.py#L14

The `callbacks` function needs to be defined in each app. To attach the callbacks to the "super" app, the function takes a Dash app as an argument and is called whenever `index.py` is run. 

I'm guessing that naming the argument to the `callbacks` function `app` is resulting in this problem. I can rename it.

tests/dashbio_demos/app_forna_container.py
shammamah: Yes, good point! I used it in this app, but I'm thinking that we should update the rest of them too. I'll open an issue for that.

tests/dashbio_demos/app_forna_container.py
alexcjohnson: ah of course, `app` is redefined down below - good catch. Yes, would be better to make the name unique inside this function.

src/lib/components/FornaContainer.react.js
```diff
@@ -18,6 +18,9 @@ export default class FornaContainer extends Component {
         this._fornaContainer = new PreFornaContainer('#' + id, {
             initialSize: [width, height],
         });
+    }
+
+    componentDidUpdate() {
         this.renderNewSequences();
```
alexcjohnson: I believe you need `renderNewSequences` in BOTH `componentDidUpdate` and `componentDidMount`, as initial render doesn't trigger `componentDidUpdate`

shammamah: Yep you're right! Fixed in 9411e5d.

src/lib/components/FornaContainer.react.js
```diff
@@ -0,0 +1,239 @@
...
+        return false;
+    }
+
+    render() {
+        return (
+            <div
+                id={this.props.id}
+                ref={ref => {
+                    this.ref = ref;
+                }}
```
alexcjohnson: FYI this works - but there's a newer way to do this with `react.createRef` https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dash-bio",
-  "version": "0.2.0",
+  "version": "0.2.1-rc4",
```
alexcjohnson: This adds a new feature, so should be released as `0.3.0`

src/lib/components/FornaContainer.react.js
```diff
@@ -0,0 +1,258 @@
...
+        const {sequences, colorScheme} = this.props;
+
+        if (!R.equals(sequences, nextProps.sequences)) {
+            return true;
+        }
+        this._fornaContainer.addCustomColors(nextProps.customColors);
+        this._fornaContainer.changeColorScheme(colorScheme);
+
+        if (nextProps.nodeFillColor !== undefined) {
+            this._fornaContainer.setOutlineColor(nextProps.nodeFillColor);
```
alexcjohnson: what happens if `props.nodeFillColor` is used, but `nextProps.nodeFillColor` is `undefined`?

shammamah: Since the `changeColorScheme` function (https://github.com/ViennaRNA/fornac/blob/9771d6d79d400e334bb306387d5fc795097ff94d/app/scripts/fornac.js#L690) is called before `setOutlineColor`, the component will default to the coloring as determined by the color scheme. The `nodeFillColor` prop is meant to override the `colorScheme` prop. 

So if `nextProps.nodeFillColor` is `undefined` but `props.nodeFillColor` is not, the component will end up being colored according to whatever is specified in `colorScheme`.

PR #420, opened 2019-09-23
Reformat generated components

mkcor:
Closes https://github.com/plotly/dash-bio/pull/411#discussion_r324820104.

## About
I am running `black` on generated components to reformat them.

## Description of changes
I am adding it to our `build:py` command.

I am happy to wait until #411 is merged...

## Before merging

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,6 +44,7 @@ jobs:
                     . venv/bin/activate
                     pip install --upgrade pip
                     pip install dash>=0.40.0
+                    pip install black
                     npm run build:py
 
 
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "lint": "eslint src --fix",
     "build:js-dev": "webpack --mode development",
     "build:js": "npm run lint & npm run format & webpack --mode production",
-    "build:py": "node ./extract-meta src/lib/components > dash_bio/metadata.json && python get_version_info.py && python -c \"import dash; dash.development.component_loader.generate_classes('dash_bio', 'dash_bio/metadata.json')\"",
+    "build:py": "node ./extract-meta src/lib/components > dash_bio/metadata.json && python get_version_info.py && python -c \"import dash; dash.development.component_loader.generate_classes('dash_bio', 'dash_bio/metadata.json')\" && black dash_bio",
     "build:all": "npm run build:js & npm run build:py",
     "build:all-dev": "npm run build:js-dev & npm run build:py"
   },
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -2,6 +2,7 @@
 -r ../requirements.txt
 
 # Additional packages needed to run the tests.
+black==19.3b0
 chromedriver-binary==2.45.0
 flake8==3.6.0
 ipdb==0.11


alexcjohnson:
You can do it this way, I suppose - I'd much rather though handle this centrally by adding black to the component generation pipeline in dash itself.

mkcor:
@alexcjohnson oh, ok, I poked @Marc-Andre-Rivet to get a sense of the consensus (or lack thereof) and I wasn't sure I should include black upstream (centrally). So, may I add black to the list of dash dependencies?

alexcjohnson:
> So, may I add black to the list of dash dependencies?

Yes, go for it! I believe we have sufficient consensus. Note that we probably *don't* want to format all .py files in the dist directory, at least not yet, since not all of them are auto-generated. You can still do that after writing each file via black's CLI, but I notice it also has a python interface, in particular [`black.format_str`](https://black.readthedocs.io/en/stable/reference/reference_functions.html#formatting) can be used before we even write the file.

mkcor:
Absolutely! I wasn't going to black any Python file other than the auto-generated ones (Python classes generated from React components). And I wanted to avoid using the CLI, so I found this [issue](https://github.com/psf/black/issues/425) when looking for a black Python interface.

I was thinking of using either [`black.format_file_in_place`](https://black.readthedocs.io/en/stable/reference/reference_functions.html#black.format_file_in_place) or [`black.reformat_one`](https://black.readthedocs.io/en/stable/reference/reference_functions.html#black.reformat_one). But your suggestion is better, it will fit right before the file writing:
https://github.com/plotly/dash/blob/75cbcd8ee133904cc4531391cee8cced0188cc5b/dash/development/_py_components_generation.py#L141
(Hmm, why isn't GitHub rendering what this permalink points to... :thinking:)

mkcor:
@alexcjohnson Oh no, "Black requires Python 3.6.0+ to run" :( it says right [here](https://black.readthedocs.io/en/stable/installation_and_usage.html). So that would be a reason to use the CLI... Or would you recommend doing some `six` magic? This [changeset](https://github.com/plotly/dash/compare/format-autogenerated?expand=1) would be so transparent...

alexcjohnson:
Ah ok, CLI it is.

mkcor:
> Ah ok, CLI it is.

Even then, it's not trivial, because we can't [install](https://github.com/plotly/dash/commit/ca3705da41c21a661aa78082d0ed66c77c4d8df4) `black` in a Python 2.7 env... So we have to `$ black <autogenerated .py files>` as an additional / separate / final step of the component generation pipeline, which can only be done from within a Python 3.6+ env, so by definition we can't fit it in the [python2.7](https://github.com/plotly/dash/blob/009c34507f95c0510287b0fe1df8ae0a2d2a89bb/.circleci/config.yml#L264) pipeline... :thinking:

alexcjohnson:
Alright, at this point I don't want to make too many concessions to Py2, but we're not ready to let it fail entirely. If we want to move forward with this I'd say we should:
- install it only where it will install, ie `black==19.3b0;python_version>="3.6"` like we do [here](https://github.com/plotly/dash/blob/75cbcd8ee133904cc4531391cee8cced0188cc5b/requires-dev.txt#L6)
- if it fails to find black, just log a warning and keep going. This may require modifying or skipping some tests in py27. Maybe then we add a new arg to `component_generator.cli` like `--strict` that will fail if it can't find black, which we can use in our own component suites' package.json scripts to prevent mis-builds.

mkcor:
Closing this PR since we want to handle this upstream.

PR #425, opened 2019-10-15
Remove alignmentviewer

### LINKED ISSUE #414: Remove and deprecate AlignmentViewer component
**Describe the bug**
Until dash-bio v0.1.3.
We have both AlignmentChart and AlignmentViewer component inside this package, which looks duplicated and causes confusion to users. 
Should look into whether one component could be deprecated and with minimum effect on backward compatibility. 

cc @rpkyle

shammamah:
Closes #414 

## About
- [ ] This is a new component
- [x] I am closing an issue

## Before merging


src/lib/components/AlignmentViewer.react.js
```diff
@@ -1,295 +0,0 @@
...
- * The Alignment Viewer (MSA) component is used to align multiple genomic
- * or proteomic sequences from a FASTA or Clustal file. Among its
- * extensive set of features, the multiple sequence alignment viewer
- * can display multiple subplots showing gap and conservation info,
- * alongside industry standard colorscale support and consensus sequence.
- * No matter what size your alignment is, Aligment Viewer is able to display
- * your genes or proteins snappily thanks to the underlying WebGL architecture
- * powering the component. You can quickly scroll through your long sequence
- * with a slider or a heatmap overview.
- * Note that the AlignmentChart only returns a chart of the sequence, while
```
ycaokris: This docs looks a bit confusing to me.Same as in Alignment Chart which we'll keep https://github.com/plotly/dash-bio/blob/447ebbe01d716a0d652e744d8fe44b781fa66ad4/src/lib/components/AlignmentChart.react.js#L17-L19

Does this `Alignment Viewer` here refers to original react component? I didn't go detailed into the difference between this two components but we could skip mentioning the one we won't be using as it'll be packed into docstring reference.

shammamah: Good catch! Will fix this ASAP :)

PR #437, opened 2019-11-08
IE11 compatibility

Marc-Andre-Rivet:
Related to https://github.com/plotly/dash/pull/1006


webpack.config.js
```diff
@@ -42,6 +42,7 @@ module.exports = {
         ],
     },
     optimization: {
+        minimize: false,
```
alexcjohnson: ?

Marc-Andre-Rivet: sry, just noise, reverting

.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ jobs:
             - run: echo $PYVERSION > ver.txt
 
             - restore_cache:
-                key: deps1-{{ .Branch }}-{{ checksum "tests/requirements.txt" }}-{{ checksum "ver.txt" }}
+                key: deps1-{{ .Branch }}-{{ checksum "requirements.txt" }}-{{ checksum "tests/requirements.txt" }}-{{ checksum "ver.txt" }}
```
Marc-Andre-Rivet: The dash version is in `requirements.txt`, updating this file during a PR causes the cache not to be updated and the changes not to be reflected in the build environment. Adding.

PR #439, opened 2019-11-18
Update instructions.

mkcor:
## About
I went through the dev setup after a period of absence.

## Description of changes
Subfolder `dash_bio/utils/` does not exist anymore, so I removed it from the CI instructions.
I updated the human instructions according to my experience re-setting up a dev env.

## Before merging


mkcor:
I'm curious how the Percy check could pass at 0f4763e but not at 8c23bf4... Is it a (tiny) non-deterministic visual change? @shammamah @byronz Do you know more?

shammamah:
@mkcor Yes, the Percy snapshots are usually inconsistent like this. This particular diff appears to be inconsequential, although I am noticing that it's not picking up on all of the snapshots taken before this particular build (since that isn't actually a new snapshot, although Percy is saying it is). Could you open an issue for this? In the meantime, I'll approve the diffs.

byronz:
pecy's algorithm of selecting reference base has a bug/or a design fault, it's not consistent.

mkcor:
@shammamah @byronz Thank you both for your answers.

> Could you open an issue for this? In the meantime, I'll approve the diffs.

I would happily do so, but I don't have a Percy account (where I could look up the details, take screenshots, link to specific visual changes...).

> Percy's algorithm of selecting reference base has a bug/or a design fault, it's not consistent.

So it's not *our* issue?

Ok, I'm waiting for your review(s)! Thanks again.

.circleci/config.yml
```diff
@@ -100,7 +100,7 @@ jobs:
                 name: Run flake8
                 command: |
                     . venv/bin/activate
-                    flake8 --max-line-length=100 tests/ dash_bio/utils/ dash_bio/component_factory/
+                    flake8 --max-line-length=100 tests/ dash_bio/component_factory/
```
shammamah: good catch! Although I'm not sure why this wasn't causing the CI tests to fail, since the `dash_bio/utils` folder doesn't even exist any more.

mkcor: I don't think this was causing any CI test to fail... unless I missed something. The entire changeset is not about fixing, only cleaning up. Thanks for you review!

PR #457, opened 2019-12-20
Update dashBio to v0.4.5 and enable asynchronous component loading

rpkyle:
This PR proposes to update the `dashBio` package for compatibility with Dash for R >= 0.2.0, which supports asynchronous component loading.

@shammamah


package.json
```diff
@@ -20,7 +20,9 @@
     "test-unit": "jest",
     "build:js-dev": "webpack --mode development",
     "build:js": "npm run lint && npm run format && webpack --mode production",
-    "build:py": "node ./extract-meta src/lib/components > dash_bio/metadata.json && python get_version_info.py && python -c \"import dash; dash.development.component_loader.generate_classes('dash_bio', 'dash_bio/metadata.json')\"",
+    "build:py": "dash-generate-components ./src/lib/components dash_bio -p package-info.json",
+    "build:py_and_r": "dash-generate-components ./src/lib/components dash_bio -p package-info.json && dash-generate-components ./src/lib/components dash_bio -p package-info.json --r-prefix='dashbio' --r-depends='manhattanly'",
+    "build:all": "npm run build:js && npm run build:js-dev && npm run build:py_and_r",
```
Marc-Andre-Rivet: Same suggestion as for DAQ
https://github.com/plotly/dash-daq/pull/79/files#r360213585

NAMESPACE
```diff
@@ -11,7 +11,6 @@ export(dashbioOncoPrint)
 export(dashbioSequenceViewer)
 export(dashbioSpeck)
 export(dashbioManhattan)
+export(dashbioAlignmentViewer)
```
shammamah: Hm, I don't think this should be here -- IIRC we got rid of the AlignmentViewer component in #425

man/dashbioAlignmentChart.Rd
```diff
@@ -25,7 +25,7 @@ width=NULL, height=NULL)
 in callbacks. The ID needs to be unique across all of the
 components in an app.}
 
-\item{eventDatum}{Named list. A Dash prop that returns data on clicking, hovering or resizing the viewer.}
+\item{eventDatum}{Character. A Dash prop that returns data on clicking, hovering or resizing the viewer.}
```
shammamah: This returns a `dict` in Python, which I think is closer to a "Named list" - @rpkyle I'm not too familiar with R, does this make sense?

package.json
```diff
@@ -43,6 +45,7 @@
     "react-oncoprint": "^1.2.3",
     "react-plotly.js": "^2.2.0",
     "react-sequence-viewer": "git://github.com/plotly/react-sequence-viewer.git",
+    "scipy": "^0.1.4",
```
shammamah: Where is this package used?

shammamah: (Maybe it was an accidental `npm install scipy` instead of `pip install scipy`)

Marc-Andre-Rivet: @rpkyle ☝️

rpkyle: Yep, that seems likely. Will fix!

man/dashbioAlignmentChart.Rd
rpkyle: @shammamah It does -- but these types are abstracted from the JavaScript source, so this isn't a change I made, but an automatic update:

https://github.com/plotly/dash-bio/blob/0b72b82609b9d036443e4cb4129e86e92504efce/src/lib/components/AlignmentChart.react.js#L47

So if it's supposed to be a `dict` or a `Named list` (in R, anyway), we'll need to make this change in JavaScript instead.

NAMESPACE
rpkyle: Hmm, strange indeed. The generator should blow away the existing `R` and `man` directories before rebuilding the `NAMESPACE` file IIRC, not sure why that didn't happen. I'll take care of this now, thanks for catching it.

rpkyle: fixed in 6e4f9c5

package.json
rpkyle: fixed in 56fe632

man/dashbioAlignmentChart.Rd
rpkyle: @shammamah Ah, now it makes sense 😉 

Have a look here: https://github.com/plotly/dash-bio/commit/96d4bbd0ed3d667ee3c55f200a63c4ff780ab62d

package.json
rpkyle: fixed in fa9eba4, f1b7c3f

man/dashbioAlignmentChart.Rd
shammamah: My bad!

rpkyle:
`package-lock.json` updated in b8c09e1

PR #459, opened 2020-01-06
refactor: deploy monorepo correctly to playground

### LINKED ISSUE #455: Set up circle-ci for monorepo-deployment
Similar as [dash-sample-apps](https://github.com/plotly/dash-sample-apps),   Change circle-ci deployment schema for batch deploy to gallery. 

This could be pooled with https://github.com/plotly/dash-bio/issues/435 which needs:

1) Remove old GA from circle-ci settings and remove them during deployment https://github.com/plotly/dash-bio/blob/74e51737830ac08da3799189ccb41d65009e0ceb/.circleci/config.yml#L142-L143

2) During deployment workflow, add a `predeploy.py`(https://github.com/plotly/dash-sample-apps/blob/master/predeploy.py) for inserting index string after `if __name__ == 'main': `
under app directory.

@josegonzalez

josegonzalez:
This change refactors the demo python files into actual standalone applications. While this does increase the overhead around copy-pasting files and keeping things in sync, it decreases the amount of work necessary to reason about what files go where.

As a side-effect, it also allows us to switch to a much faster deploy process because we don't need to clone the entire repository.

Finally, the GA tag is updated to the latest version.

Closes #455

## About

- [x] I am closing an issue

## Before merging
  - Do I need to do this?
  - This changes the deploy process so I think so :)


tests/dashbio_demos/dash-alignment-viewer/app.py
```diff
@@ -552,7 +549,7 @@ def layout():
         ]),
         dcc.Loading(className='dashbio-loading', children=html.Div([
             dash_bio.AlignmentChart(
-                id='alignment-chart',
+                id='alignment-viewer',
```
shammamah: These should all be changed to `alignment-chart`. The name of the app is actually wrong -- it should be `app_alignment_chart.py` (or, in the new structure, `dash-alignment-chart/app.py`.

tests/dashbio_demos/dash-alignment-viewer/assets/base-app-page.css
```diff
@@ -0,0 +1,248 @@
+@import url('https://fonts.googleapis.com/css?family=Open+Sans:300,400');
```
shammamah: Is there a way to keep this at a higher level within the directory and then copy it into the folder containing the app that's being deployed? I think that might be better than having to copy/paste this file into every single app's folder.

tests/dashbio_demos/dash-alignment-viewer/layout_helper.py
```diff
@@ -0,0 +1,109 @@
+import base64
```
shammamah: This file should also be added to the app directory *at* the deployment step instead of having copies of it in every single app's folder. Maybe a CircleCI command to copy this file into the folder that is being deployed?

tests/dashbio_demos/dash-alignment-viewer/assets/base-app-page.css
josegonzalez: I don't know if the standalone `index.py` at the repo root will work properly if we do that. I'll take a crack at it though, since I agree, the copy-pasta doesn't feel great.

tests/dashbio_demos/dash-alignment-viewer/layout_helper.py
josegonzalez: The problem here is that you cannot do relative imports above the path you are in, so moving this up  will _definitely_ break the `index.py`. If we are fine with getting rid of that fake gallery index, then this is trivial, but if not, this won't do.

Also, this will make it harder for folks to copy out the sample app into their own apps since they'll need to remember to copy this file from a different place. Probably not a huge deal since this was already necessary.

tests/dashbio_demos/dash-alignment-viewer/app.py
josegonzalez: Should I rename the app in the gallery as well?

shammamah: It should already be renamed! Looks like I must have forgotten to change the selectors and the app title that's displayed in the header -- my mistake. https://dash-gallery.plotly.host/dash-alignment-chart/

tests/dashbio_demos/dash-alignment-viewer/assets/base-app-page.css
shammamah: We could probably get rid of that. It's really only around for heroku test deployments and I personally haven't looked at it in a long time now that we have integration/visual tests. CC @mkcor since she would have some more valuable input into this issue!

mkcor: One `assets/` folder per demo app looks tedious indeed... but I can see how it might be necessary to make each app actually standalone... I'm not sure, I would need to dig deeper.

ycaokris:
insertion of GTM looks good.  I'll check after gallery deploy is finished  👍

tests/dashbio_demos/dash-alignment-viewer/app.py
josegonzalez: Fixed!

josegonzalez:
@mkcor @shammamah I can refactor this to remove the index.py and duplicate files and then someone can disable heroku auto-deploys. Given that we will autodeploy to the playground for all PRs and then the gallery on the `master` branch, I think that seems fine. Thoughts?

shammamah:
@josegonzalez That sounds good to me!

ycaokris:
@josegonzalez  I've moved this into solutions project board for following up GA tag issue until gallery apps are updated 👍 .

josegonzalez:
I'm guessing percy does testing based on the heroku app, which should be disabled since I removed the index.py from this codebase.

shammamah:
@josegonzalez Percy doesn't use Heroku -- all the snapshots are taken in our integration tests.

josegonzalez:
Gotcha. I don't have access to the percy org, so I can't check on the change. I'll leave that one up to whomever reviews :)

shammamah:
Just approved :) I'll remove the Heroku pipeline, since we won't be needing it any more. Could you also make a change to the PR template by adding a line that asks the user to specify the URL of the app(s) that they're changing? (Should be here: https://github.com/plotly/dash-bio/blob/master/.github/pull_request_template.md)

Speaking of the PR template, it looks like the "steps to take before merging" and some other parts of it are outdated -- could you open up an issue and assign it to me to ensure that gets updated?

josegonzalez:
Note: I can't merge since there is a pending (but fixed) change from @ycaokris. I'd imagine I still might not be able to merge due to repo perms.

tests/dashbio_demos/dash-alignment-chart/app.py
```diff
@@ -702,9 +699,9 @@ def customize_showid(val):
 
 
 # only declare app/server if the file is being run directly
-if 'DASH_PATH_ROUTING' in os.environ or __name__ == '__main__':
+if 'DEMO_STANDALONE' not in os.environ:
     app = run_standalone_app(layout, callbacks, header_colors, __file__)
     server = app.server
 
-if __name__ == '__main__':
+if __name__ == "__main__":
```
shammamah: If possible, could you please change these back to single quotes (since this diff isn't relevant to the PR)?

PR #460, opened 2020-01-14
💄 Spice up the CircleCI config file

jbampton:


mkcor:
Personally I'm not into that... :thinking:

Marc-Andre-Rivet:
@jbampton Thanks for the suggestion. We do use something similar in DCC because the dev who updated the config file was into it but we're not looking to make this a standard. We'll be passing on this one.

PR #463, opened 2020-01-16
Let CI run Python unit tests

### LINKED ISSUE #462: CI does not run Python unit tests.
**Describe the bug**
The current version of our CircleCI config does not account for Python unit tests (to run, say, with `pytest`). That's probably because, up until #458, there were no such tests.

**To Reproduce**
Push a feature branch to `origin` and notice that Python unit tests in `./tests/unit/` are not run.

**Expected behavior**
Python jobs for 7a094515ddeb4637c529ff6fbcbcf708e60f7046 should have failed.

**Screenshots**
python-unit-tests

**Python 3.6 or 3.7**

mkcor:
Closes #462.

## About
- [ ] This is a new component
- [x] I am closing an issue

## Description of changes
~I was quite tempted to add a Python command right next to `python unit_test_data_setup.py`. Instead, should we separate Python unit tests and JS unit tests into two different steps?~
Oops, that step of running Python unit tests definitely can't be part of a `node` job! I have rebased.

## Before merging

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,6 +110,13 @@ jobs:
                     pip install -e .
                     pytest tests/integration
 
+            - run:
+                name: Run unit tests
+                command: |
+                    python -m venv venv
+                    . venv/bin/activate
+                    pytest tests/unit
+
     python-3.7:
         <<: *test-template
         docker:


shammamah:
>should we separate Python unit tests and JS unit tests into two different steps?

I don't think so. The unit tests are meant to test components individually. Some of those components are written in Python instead of JS, but that doesn't mean they're two different types of tests.

mkcor:
> > should we separate Python unit tests and JS unit tests into two different steps?
> 
> I don't think so. The unit tests are meant to test components individually. Some of those components are written in Python instead of JS, but that doesn't mean they're two different types of tests.

Oops, I have rebased and edited the description (see above) before seeing your comment... I agree they are both unit tests, but since the CI config is structured into `node` vs `Python` jobs, jobs being two levels above steps... it feels wrong not to have this `pytest` command under the Python jobs (i.e., 'where it belongs').

mkcor:
@shammamah by the way ;)

> The unit tests are meant to test components individually.

This is not correct. Unit tests are meant to test *code units* (such as functions). The component level does not come into play; for convenience, we can use it in the file structure of course (e.g., file `tests/unit/test_clustergram.py` will gather tests for code units found in the Clustergram component). But the actual tests bear on code units.

PR #443, opened 2019-11-21
Rebuild package.

mkcor:
## About
I ran
```bash
npm run build
npm update
npm audit fix
```

## Description of changes
I'm not sure why the `./dash_bio/*.js` files changed as a result... But some catching up must be going on.
I fixed the docs following the [renaming of `build:all`](https://github.com/plotly/dash-bio/commit/a1722b1a4993edef37b4de0a5faefe84c3d6b649), @Marc-Andre-Rivet.
I documented the change we just merged over at #423.

## Before merging
- [ ] I have gone through the [code review checklist](https://github.com/plotly/dash-component-boilerplate/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/review_checklist.md)
- [ ] I have completed all of the [steps to take before merging](https://github.com/plotly/dash-bio/blob/master/.github/before_merging.md)
- [ ] I know how to [deploy to DDS](https://github.com/plotly/dash-bio/blob/master/.github/deploying_to_dds.md)

PR #480, opened 2020-02-09
Update eslint to version 6 (major).

mkcor:
## About
This PR is about package maintenance. Ever since #443, I've been thinking about updating our Node dependencies, at least to new minor versions (i.e., with non-breaking changes) that are now available.

## Description of changes
Regarding `eslint`, I went for a major upgrade (from version 5 to 6), so that we don't stay behind and we move away from `es5` (https://github.com/shipshapecode/shepherd/issues/537). Since we are talking about a linter, this update shouldn't change any functionality in Dash Bio.

For reference, I ran the following sequence of commands:
```sh
npm ci
npm outdated
npm update
npm outdated
npm install eslint@6.8.0
npm install eslint-config-prettier@6.10.0
# edit postbuild command to use es6 instead of es5
npm ci
npm run build

```


mkcor:
@shammamah please review!

alexcjohnson:
Our source files should use and lint for es6, but the bundles should still be transpiled to es5 and checked as such for the foreseeable future.

mkcor:
Ok, thanks for the pointers.

So I'll revert
```
"postbuild": "es-check es6 dash_bio/*.js",
```
to
```
"postbuild": "es-check es5 dash_bio/*.js",
```
(and rebuild of course) but... I'll keep the eslint upgrade from 5 to 6 (`"eslint": "^6.8.0",`), right? So that our source files lint for es6 (via `eslint src`).

mkcor:
Isn't this necessarily incompatible? As mentioned point 1 of https://github.com/plotly/dash/pull/1006#issue-338281994: "The artifacts are now ES5-validated with es-check, making it impossible to get a successful build with hanging ES6+ syntax" so this error is expected:
```
> es-check es5 dash_bio/*.js

ES-Check: there were 1 ES version matching errors.

          ES-Check Error:
          ----
          · erroring file: dash_bio/async~moleculeviewer3.js
          · error: SyntaxError: The keyword 'const' is reserved (1:132708)
          · see the printed err.stack below for context
          ----
```
I'm happy to downgrade eslint back to 5.12.1, but then how can we ensure our source files use and lint for es6, @alexcjohnson?

alexcjohnson:
I don't see why the eslint version would have anything to do with what ends up in our bundles. My suspicion is something changed about the `3dmol` package, which was implicitly updated here from v1.3.7 to v1.5.0, such that the files we're pulling in are now using es6. Normally files in `node_modules` are skipped by the transpilation step, but we already have to re-enable it for a few packages:

https://github.com/plotly/dash-bio/blob/9523a2ce2b803c1a9779580fb49b758fa723aa9f/webpack.config.js#L26

Probably need to add `3dmol` to that list as well.

mkcor:
@alexcjohnson I thought that the JS files under `src/` came into play as inputs for `npm run build` but I see that `build:js` lints, formats, and 'webpacks' where only `index.js` does (come into play):
https://github.com/plotly/dash-bio/blob/9523a2ce2b803c1a9779580fb49b758fa723aa9f/webpack.config.js#L9
Thanks for pointing this config to my attention!

mkcor:
@shammamah @alexcjohnson PR ready for final review :)

PR #482, opened 2020-02-19
Replace all instances of plot.ly with plotly.com.

shammamah:
### About 
Related to https://github.com/plotly/dash-core/issues/152


Marc-Andre-Rivet:
@shammamah Needs update

PR #489, opened 2020-02-27
Issue 488 - Rename async modules

### LINKED ISSUE #488: Rename async modules
Same as https://github.com/plotly/dash-core-components/issues/745

Marc-Andre-Rivet:
Closes #488


.circleci/config.yml
```diff
@@ -103,11 +100,11 @@ jobs:
...
                 command: |
                     . venv/bin/activate
-                    pip install --upgrade pip
-                    pip install -r tests/requirements.txt
                     npm ci
                     npm run build
-                    pip install -e .
+                    python setup.py sdist
+                    cd dist
+                    find . -name "*.gz" | xargs pip install --no-cache-dir --ignore-installed && cd ..
```
Marc-Andre-Rivet: Using the opportunity to install from tarball instead of the local folder `-e .`. This change helps identify issues when resource paths change and the MANIFEST.in file isn't synced.

.circleci/config.yml
```diff
@@ -76,9 +74,8 @@ jobs:
                 name: Install requirements
                 command: |
                     . venv/bin/activate
-                    pip install --upgrade pip
-                    pip install -r tests/requirements.txt --quiet
-                    pip install dash[testing]
+                    pip install --progress-bar off -r tests/requirements.txt --quiet
+                    pip install --progress-bar off dash[testing]
```
Marc-Andre-Rivet: No point having thousands of lines of progress bar on each CI run..

.circleci/config.yml
```diff
@@ -103,11 +100,11 @@ jobs:
                 name: Run integration tests
                 command: |
                     . venv/bin/activate
-                    pip install --upgrade pip
-                    pip install -r tests/requirements.txt
```
Marc-Andre-Rivet: This step is run after the step above that already installs everything, 🔪

shammamah: Good catch :)

requirements.txt
```diff
@@ -1,14 +1,14 @@
...
--e .
+
 cython>=0.19
 dash>=1.6.1
 dash-bio-utils==0.0.4
 dash-daq==0.2.2
 gunicorn==19.9.0
 jsonschema==2.6.0
 matplotlib==3.0.2
-numpy==1.15.4
```
alexcjohnson: I feel like I've asked this question before, but does this root `requirements.txt` file get used anywhere?

shammamah: This was originally used for the demo apps but I don't think it needs to be here any more after https://github.com/plotly/dash-bio/pull/459 since the apps each have their own `requirements.txt`. We should probably replace it with a `dev-requirements.txt` file instead.

alexcjohnson: @shammamah Good to know, thanks! No need to fix now but please make an issue for followup.

Issue #451, opened 2019-11-24
Make plots work with pandas dataframes as input


joelostblom:
From the documentation, it seems like some dashbio plots (e.g. volcano and manhattan) already work with pandas dataframes, while others (e.g. clustergram) does not. I haven't looked through all the plots, so I don't know how generalizable this suggestion is, but at least for clustergram plots it would be helpful to pass data frames and string references to dataframe columns as inputs.

As an example, instead of writing

```python
db.Clustergram(iris.select_dtypes('number').to_numpy(), row_labels=iris['species'].to_list())
```

it would be more intuitive to write

```python
db.Clustergram(iris, row_labels='species')  # or a reference to a column with color names
```

mkcor:
Hi @joelostblom,

Once #478 is reviewed, approved, and merged, you will be able to run `db.Clustergram(iris)`!

I will troubleshoot #449 before passing the DataFrame's labels to `column_labels` and/or `row_labels` whenever these arguments aren't specified.

PR #492, opened 2020-02-28
Fix Clustergram when passing labels.

### LINKED ISSUE #449: Clustergram row labels change heatmap values/colors
Passing a list of labels to `row_labels` seem to change the values that are plotted in the heatmap rather than just adding a textual label on the side.

```python
import dash_bio as db
import plotly.express as px

iris = px.data.iris()
db.Clustergram(iris.select_dtypes('number').to_numpy())
```

image

In the plot above there it one value per row in each color. Below, rows are grouped together so that there are are only three values/colors per column.

```python
db.Clustergram(iris.select_dtypes('number').to_numpy(), row_labels=iris['species'].to_list())
```

image

The font size of the labels also don't adjust to the size of the plot. I have to change the height to 2000 before I can read what they say:

image

```
-----
dash      	1.6.1
dash_bio  	0.4.4
dash_core_components	1.5.1
dash_html_components	1.0.2
numpy     	1.17.3
pandas    	0.25.3
plotly    	4.3.0
-----
IPython   	7.9.0
jupyter_client	5.3.3
jupyter_core	4.6.1
notebook  	6.0.1
-----
Python 3.8.0 | packaged by conda-forge | (default, Nov 22 2019, 19:11:38) [GCC 7.3.0]
```

mkcor:
Fixes #449 

## About
- [x] I am closing an issue
- [ ] This is a new component

## Description of changes

When clustering along rows (resp. columns), the bug would appear for non-unique row (resp. column) labels mixed with non-unique rows (resp. columns). For example, it would get unnoticed when calling `Clustergram(data, row_labels=row_labels)` if
```py
> data
array([[1, 1, 1, 1],
       [3, 3, 3, 3],
       [1, 1, 1, 1],
       [3, 3, 3, 3],
       [1, 1, 1, 1],
       [3, 3, 3, 3]])
```
and
```py
> row_labels
['a', 'b', 'a', 'b', 'a', 'b']
```
but the clustering would fail if, instead,
```py
> row_labels
['a', 'b', 'b', 'b', 'b', 'b']
```
(as used in the unit tests), leading to the following (wrong) clustering:
```py
array([[3, 3, 3, 3],
       [1, 1, 1, 1],
       [3, 3, 3, 3],
       [3, 3, 3, 3],
       [3, 3, 3, 3],
       [3, 3, 3, 3]])
```

Apparently I forgot to push after I committed 3c83048; I wanted to showcase the (wrong) clustering result (along columns):
```py
ipdb> curves_dict['heatmap']['z']                                                                         
array([[3, 1, 3, 3, 3, 3],
       [3, 1, 3, 3, 3, 3],
       [3, 1, 3, 3, 3, 3],
       [3, 1, 3, 3, 3, 3]])
ipdb> clustered_data                                                                                      
array([[1, 1, 1, 3, 3, 3],
       [1, 1, 1, 3, 3, 3],
       [1, 1, 1, 3, 3, 3],
       [1, 1, 1, 3, 3, 3]])
```
before implementing the fix (f23f7ed).


mkcor:
I noticed that we didn't have any integration test for prop `hidden_labels` so I added one with cf5c66ba3c51765f14e25c06d1615a680b2f3aec. While running the Clustergram integration tests locally, I noticed that (row or column) labels would *not* show up in the (testing) 'simple app' although I passed them via, e.g.,
```py
dash_bio.Clustergram(
    data=data,
    row_labels=row_labels,
    column_labels=col_labels
)
```
e0a60eb149660493949c92446f7c8097d7c35676 fixed it.

With the latter change, I also meant to echo what @joelostblom suggested [at some point](https://github.com/plotly/dash-bio/issues/451#issue-527668796), i.e., (whenever input data is a dataframe) we should use the dataframe's labels by default: Say `df` is the input dataframe, `df.index` (resp. `df.columns`) get passed as `row_labels` (resp. `column_labels`) by default.

Note that, in this spirit, the Clustergram [demo app](https://dash-gallery.plotly.host/dash-clustergram/) uses prop/parameter `row_labels_source` (which is not readily available in the Clustergram component).

Finally, this change (e0a60eb149660493949c92446f7c8097d7c35676) is 'non-breaking' even if it lays the ground for a new behaviour. To give all the details, it does not affect the current rendering in terms of *not showing* row (resp. column) labels whenever `row_labels` (resp. `column_labels`) is not specified, even if `hidden_labels=['row']` (resp. `hidden_labels=['col']`) is not specified either. Indeed:

* we are now skipping
https://github.com/plotly/dash-bio/blob/80693ec9e95c1cf234d1e41efc6b3209cf340cf7/dash_bio/component_factory/_clustergram.py#L248
* and hence
https://github.com/plotly/dash-bio/blob/80693ec9e95c1cf234d1e41efc6b3209cf340cf7/dash_bio/component_factory/_clustergram.py#L318
* and hence
https://github.com/plotly/dash-bio/blob/80693ec9e95c1cf234d1e41efc6b3209cf340cf7/dash_bio/component_factory/_clustergram.py#L535
but since `row_labels is None`, so is `self._row_labels` here 
https://github.com/plotly/dash-bio/blob/80693ec9e95c1cf234d1e41efc6b3209cf340cf7/dash_bio/component_factory/_clustergram.py#L526
(printing nothing then).

mkcor:
I'm not sure which version I should update to: The latest changes to CHANGELOG by @Marc-Andre-Rivet introduced a `## [Unreleased]` section with two changes, so I'm not sure the next version should be patch (`## [0.4.8] - `) or minor (`## [0.5.0] - `).

mkcor:
@shammamah I added 1fc32b4 assuming the build would be fine and we could release today, but apparently it's not the case! I went like
```
npm ci
npm run build
npm i  # to update package-lock.json
npm audit fix  # following npm's suggestion
```
@Marc-Andre-Rivet would you mind having a look? Thanks!

mkcor:
By the way, what happened to the CI visual (Percy) check?

shammamah:
@mkcor 

There is nothing in `package-lock.json` that should be updated in this PR (other than the version number), since it's updating a Python component. Please refer to the [contributing guidelines](https://github.com/plotly/dash-bio/blob/master/CONTRIBUTING.md#installing-new-npm-packages): `npm i` should only be run when installing a new npm package (which you would do if working on a React component).

Please revert the changes and only run `npm ci; npm run build` after updating `package.json`.

mkcor:
Thanks for the quick response, @shammamah ! The changes to `package-lock.json` other than the version number come from running `npm audit fix` indeed.

But the version number change for `dash-bio` in `package-lock.json` was obtained from running `npm i` -- how else can we achieve the package version update in `package-lock.json`?

shammamah:
@mkcor If you just run `npm i` after `npm ci; npm run build` it should update `package-lock.json` in the way that we would expect (just the version of the package itself gets changed)! I think the only issue here was the `npm audit fix`. 

Re: the tests, we're seeing something similar in other Dash repos (e.g., [here](https://github.com/plotly/dash-core-components/pull/768)). This might have to do with the `pytest`-related changes made in https://github.com/plotly/dash/pull/1151.

mkcor:
@shammamah exactly, that's what I did. So I _am_ allowed to run `npm i`? ^^

mkcor:
Done. It's just a little stressful to let
```
found 37991 vulnerabilities (37850 low, 136 moderate, 5 high)
  run `npm audit fix` to fix them, or `npm audit` for details
```
be, with 'high' printed in red... :s 

Thanks for pointing me to related Python test issues!

mkcor:
@shammamah @Marc-Andre-Rivet please review/approve: Could we release today? Thanks!

Issue #467, opened 2020-01-20
merged pdb structure: any protein + DNA (1BNA.pdb) breaks Molecule3D


IvoLeist:
**Describe the bug**
If I am trying to upload a merged pdb structure (any protein + DNA (1BNA.pdb) 
Molecule3D stops working. This holds true dash-bio on localhost and
https://dash-gallery.plotly.host/dash-molecule-3d-viewer/

**To Reproduce**

1. Download two pdb structures from rcsb.org
     (e.g. https://files.rcsb.org/download/1BNA.pdb and https://files.rcsb.org/download/2MRU.pdb)
2. Upload them in pymol (https://pymol.org/2/) 
3. Save them in a new pdb file
4. Upload this file

Or download the attached example (not yet attached)

**Expected behavior**
It should show the two molecules

**Bug message if hosted on local host**
will add it in a minute

IvoLeist:
Below is the error showing up on 
https://dash-gallery.plotly.host/dash-molecule-3d-viewer/

The number in front of Uncaught TypeError: Cannot read property 'children' of null
is constantly increasing + the app cannot be used any longer 
(e.g. selecting a new molecule from the dropdown or uploading a new one.
*Page has to be reloaded to use it again

![image](https://user-images.githubusercontent.com/28399610/72728022-abae2b80-3b8c-11ea-88a5-afe4370c3072.png)

PR #496, opened 2020-03-05
add dash-ngl

IvoLeist:
## About
- [ ] I am closing an issue
- [X] This is a new component

## Description of changes
as discussed with @rpkyle he wanted me to do a PR to add my
dash component https://github.com/IvoLeist/dash_ngl
to dash-bio to replace moleculeViewer3d

It is inspired by the shortcomings of https://dash.plot.ly/dash-bio/molecule3dviewer which does not support .cif files and has issues in showing a merged pdb structure: any protein + DNA https://github.com/plotly/dash-bio/issues/467

## Features
- [x] Protein selection by dropdown
- [x] Support for .pdb (.gz) and .cif(.gz) files
- [x] Show multiple structures side by side
- [x] Filter the structure to show only one chain (e.g. 6CHG.A)
- [x] Filter the structure to show only a residues range of one chain (e.g. 6CHG.A:1-450)
- [x] Changeable viewer background (black / white)
- [x] Changeable camera settings ( perspective / orthographic)
- [x] Changeable chain colors
- [x] Changeable molecular representations (cartoon, sticks etc.)
- [x] Highlight of chosen Atoms (represented as balls)
- [x] Highlight C alpha of chosen residues (represented as balls)
- [x] Download of the structure as png

<p align="center">
  
</p>

## Limitations
~So far only 2-3 structures at the same time are supported~
- A structure can only be filtered to one chain
- Placement and zooming is far from perfect
~zipped files are not (yet) supported~


package.json
```diff
@@ -45,7 +45,8 @@
     "react-oncoprint": "^1.2.3",
     "react-plotly.js": "^2.4.0",
     "react-sequence-viewer": "git://github.com/plotly/react-sequence-viewer.git",
-    "speck": "git://github.com/plotly/speck.git"
+    "speck": "git://github.com/plotly/speck.git",
+    "ngl": "^0.10.4"
```
shammamah: Could you please move this so that it is in alphabetical order?

src/lib/components/DashNgl.react.js
```diff
@@ -0,0 +1,253 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import {Stage, Selection} from 'ngl';
+
+/**
+ * Dash ngl component
```
shammamah: Please include a longer description here - you can take a look at the other components for some inspiration.

src/lib/components/DashNgl.react.js
```diff
@@ -0,0 +1,253 @@
...
+ */
+export default class DashNgl extends Component {
+    // constructor might not be needed anylonger:
+    // https://hackernoon.com/the-constructor-is-dead-long-live-the-constructor-c10871bea599
+    constructor(props) {
+        super(props);
+        this.state = {stage: null, structuresList: []};
+        console.log(this.props);
```
shammamah: Please remove these calls to `console.log`! There are quite a few other ones in the file.

src/lib/components/DashNgl.react.js
```diff
@@ -0,0 +1,253 @@
...
+
+        this.setState({stage});
+        console.log('component did mount');
+    }
+
+    // triggered by any update of the DOM (e.g. new dropdown selection)
+    shouldComponentUpdate(nextProps) {
+        const {stageParameters} = this.props;
+
+        // when the app is starting data is not correctly interpreted as an array
```
shammamah: Could you elaborate on this? Is `this.props.data` `undefined`, or is it something else?

src/lib/components/DashNgl.react.js
```diff
@@ -0,0 +1,253 @@
...
+        // but after a few seconds it is therefore this if-else statement
+        if (Array.isArray(this.props.data) === false) {
+            return false;
+        }
+        const oldSelection = this.props.data[0].selectedValue;
+        const newSelection = nextProps.data[0].selectedValue;
+
+        // check for stage params changed
+        if (stageParameters !== nextProps.stageParameters) {
+            // console.log("stage params changed")
```
shammamah: Please get rid of these commented lines.

src/lib/components/DashNgl.react.js
```diff
@@ -0,0 +1,253 @@
...
+
+    /**
+     * CSS styling for viewport container
+     */
+    viewportStyle: PropTypes.object,
+
+    /**
+     * Parameters for the stage
+     */
+    stageParameters: PropTypes.object,
```
shammamah: Please be a bit more specific here - this should be a `PropTypes.exact` instead. It would also be great to have comments that are a bit more descriptive and in full sentences, since those are what go in the Python docstring. You can see this component for an example of how it should look: https://github.com/plotly/dash-bio/blob/dcdd7161c3a13790f2b6c6ca26ddad7eed648225/src/lib/components/FornaContainer.react.js#L202

tests/dashbio_demos/dash-ngl/app.py
```diff
@@ -0,0 +1,238 @@
...
+def layout():
+    return html.Div(
+        [
+            # header
+            html.Div(
+                children=[html.H1("NGL Protein Structure Viewer")],
+                style={"backgroundColor": "#3aaab2", "height": "7vh"},
+            ),
+            viewer,
+            # if you use dcc.Loading viewer the app goes into a mount loop!
```
shammamah: Could you please elaborate on this?

tests/dashbio_demos/dash-ngl/app.py
```diff
@@ -0,0 +1,238 @@
...
+            # dcc.Loading(viewer),
+            dcc.Tabs(
+                id="tabs",
+                children=[
+                    dcc.Tab(
+                        label="Data",
+                        children=[
+                            # daq.ToggleSwitch(
```
shammamah: Please get rid of these commented lines!

shammamah:
Please also make sure to run `pylint` on the app that you've created - that will get rid of quite a few of those Python CI test failures! :)

IvoLeist:
Just added another commit adressing @shammamah 's concern.
Thank you for the quick first review and the positive feedback :)

tests/dashbio_demos/dash-ngl/app.py
IvoLeist: Interesting this bug magically vanished at least on my laptop (Brave)*
I am going to add `dcc.Loading` in another commit.

*It might still not work on my office machine (Chromium)
going to try that on Monday

src/lib/components/DashNgl.react.js
IvoLeist: That is a weird bug if you console.log (this.props.data) during the start of the app its properties are changing. See in the last line in the picture below
it somehow changed to true (=array). Since I did not found any explanation for that behaviour I "fixed" that error with this if-else statement.

Do you have a clue why this is happening ?
```
    32         const {id, stageParameters} = this.props;
    33         console.log(this.props)
    34         console.log(this.props.data)
    35         console.log(typeof this.props.data)
    36         console.log(Array.isArray(this.props.data))
```
console log_thisProps

IvoLeist:
@shammamah I just finished my first python test :)
Is it possible to test it locally or should I just push it ?

shammamah:
@IvoLeist That's great to hear! You should be able to run it locally with `pytest -k tests/integration/yourtest.py`. Make sure that you check out the testing portion of the `CONTRIBUTING.md` if you get stuck: https://github.com/plotly/dash-bio/blob/master/CONTRIBUTING.md#step-3-run-tests-locally

src/lib/components/DashNgl.react.js
```diff
@@ -225,25 +184,32 @@ DashNgl.defaultProps = {
...
-     * The ID used to identify this component in Dash callbacks.
+     * The ID of this component, used to identify dash components in
+     * callbacks. The ID needs to be unique across all of the
+     * components in an app.
      */
     id: PropTypes.string,
 
     /**
-     * CSS styling for viewport container
+     * The height (in px) and the width (in %) of the container
```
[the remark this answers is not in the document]
IvoLeist: This is is on purpose because I did not manage to get an automated resizing of the width of the ngl viewer using pixels. But this whole section has to be improved any way...
I am actually not using that variable but hard coding the values in the render function because somehow the render function does not see `const defaultViewportStyle`

In my component it works https://github.com/IvoLeist/dash_ngl
Do you have any what is the reason for that ?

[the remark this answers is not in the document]
shammamah: You could try using the Chrome debugger to read the values of `height` and `width` whenever the `render` function runs (or you could also just print it). My guess is that you're setting the height and width somewhere in your demo app, so when the initial callback gets fired, it sends a `None` value which might override the defaults.

IvoLeist:
@shammamah sorry to bother you again but unfortunately I am still stuck with local testing...
```
pip install dash[testing]
zsh: no matches found: dash[testing]
``

**Comment by shammamah** [comment]
@IvoLeist No worries! `pip` might be installing packages to a Python 2 environment; could you try again with `pip3 install dash[testing]`? (And make sure you're on the correct version of Python by running `python --version`)

**Comment by IvoLeist** [comment]
❯ python --version
Python 3.7.6
❯ pip3 install dash[testing]
zsh: no matches found: dash[testing]

**Comment by shammamah** [comment]
Ah, I see what's going on! Try escaping the square brackets: https://stackoverflow.com/questions/30539798/zsh-no-matches-found-requestssecurity

So run 
```
pip install dash\[testing\]
```

**Comment by IvoLeist** [comment]
@shammamah thanks for the quick google.
I had no idea about this issue regarding zsh (recently switched to it)

Now I almost there :)
Now it fails because it does not find dash_bio ?

> ―――――――――――――――――――――――――――――――――――――――――――――― tests/integration/test_volcano_plot.py ――――――――――――――――――――――――――――――――――――――――――――――
> ImportError while importing test module '/home/ivo/projects/bioinfo/repos/dash-bio/tests/integration/test_volcano_plot.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> tests/integration/test_volcano_plot.py:6: in <module>
>     import dash_bio
> E   ModuleNotFoundError: No module named 'dash_bio'
>

**Comment by shammamah** [comment]
This might be due to the `pip/pip3` thing I mentioned earlier. Could you try `pip3 install dash-bio`?

**Comment by IvoLeist** [comment]
@shammamah ah with this pip/pip3 it is so confusing :D
Why do you not support conda (?)

The local test run in 54 errors I guess all related to the selenium driver
I guess I have to figure out how to install it.

**Comment by IvoLeist** [comment]
@shammamah update chromedriver was easy to get
*I love Aur :dancer: 

But there is unfortunately a new error...

> ―――――――――――――――――――――――――――――――――――――――――――――――――― tests/unit/test_clustergram.py ――――――――――――――――――――――――――――――――――――――――――――――――――
> import file mismatch:
> imported module 'test_clustergram' has this __file__ attribute:
>   /home/ivo/projects/bioinfo/repos/dash-bio/tests/integration/test_clustergram.py
> which is not the same as the test file we want to collect:
>   /home/ivo/projects/bioinfo/repos/dash-bio/tests/unit/test_clustergram.py
> HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
>

**Comment by shammamah** [comment]
@IvoLeist Oops, I misled you... you should just be running 
```
pytest tests/integration/test_dashNgl.py
```

The `-k` option means that you're specifying a keyword, so it makes sense that `pytest` doesn't know whether to run the unit test or the integration test for `clustergram`: https://docs.pytest.org/en/latest/usage.html

Sorry about that - could you just try running `pytest` without the `-k` option and let me know if that works?

File: tests/dashbio_demos/dash-ngl/app.py [hunk 10]
```diff
@@ -0,0 +1,223 @@
...
+        "ext": ext,
+        "config": {"type": "text/plain", "input": contents},
+    }
+
+
+def callbacks(_app):
+
+    # Callback for molecule visualization based on the dropdown selection
+    @_app.callback(
+        [Output("viewport", "data")],
```
shammamah:
```suggestion
        Output("viewport", "data"),
```
Very subtle error here -- if you wrap `Output` in a list like you've done here, Dash is expecting you to return a list or tuple of values from the callback that have a one-to-one correspondence with the items in the list. I see that `data` is also a list, but it's actually the value you want to *return* -- so in this case, you would have to return something like `[data]` or `(data,)`. It's easier (and makes more sense) to just drop the square brackets here.

shammamah:
Re: this: 
>Why do you not support conda (?)

The best answer I can give you is that we just haven't had the time/resources to seriously look into it. We do know that a lot of people use conda, though!

src/lib/components/DashNgl.react.js
```diff
@@ -0,0 +1,219 @@
...
+    constructor(props) {
+        super(props);
+        this.state = {stage: null, structuresList: []};
+    }
+
+    componentDidMount() {
+        const {id, stageParameters} = this.props;
+        const params = {...stageParameters};
+
+        const stage = new Stage(id, params);
```
shammamah:
```suggestion
        const {id, stageParameters, viewportStyle} = this.props;
        const params = {...stageParameters};

        const stage = new Stage(id, params);
        stage.setSize(viewportStyle.width, viewportStyle.height);
```

src/lib/components/DashNgl.react.js
```diff
@@ -0,0 +1,219 @@
...
+                        data.filename,
+                    ]),
+                }));
+            });
+    }
+
+    render() {
+        return (
+            <div>
+                <div id="viewport" style={{width: '100%', height: '500px'}} />
```
shammamah:
```suggestion
                <div id="viewport" />
```

src/lib/components/DashNgl.react.js
```diff
@@ -0,0 +1,219 @@
...
+     * components in an app.
+     */
+    id: PropTypes.string,
+
+    /**
+     * The height (in px) and the width (in %) of the container
+     * in which the molecules will be displayed.
+     * It should be in JSON format
+     */
+    viewportStyle: PropTypes.object,
```
shammamah:
```suggestion
    viewportStyle: PropTypes.exact({
        width: PropTypes.string,
        height: PropTypes.string
    })
```

src/lib/components/DashNgl.react.js
```diff
@@ -0,0 +1,219 @@
...
+        return (
+            <div>
+                <div id="viewport" style={{width: '100%', height: '500px'}} />
+            </div>
+        );
+    }
+}
+
+const defaultViewportStyle = {
+    width: '100%',
```
shammamah:
```suggestion
    width: '500px',
```

src/lib/components/DashNgl.react.js
```diff
@@ -101,36 +77,28 @@ export default class DashNgl extends Component {
...
         stage.animationControls.moveComponent(stageObj, stageObj.getCenter());
-        // stage.autoView()
     }
 
-    // If user has selected structure already just add the new Representation
+    // If user has selected structure load it from the browser
     loadStructure(stage, filename, chain, color, xOffset) {
-        console.log('load from browser');
-        // console.log(filename)
         const stageObj = stage.getComponentsByName(filename).list[0];
```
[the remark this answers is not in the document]
IvoLeist: filename is literally the filename how it is named in the data folder (e.g. 6CHG.pdb). It is taken from the information coming from the backend to the frontend. These are collected in `componentDidUpdate`:
`const {data, stageParameters} = this.props;`

And the filename is attached in the loadData method to the stageObj see code snippet below
```
loadFile(stringBlob, {ext: data.ext, defaultRepresentation: false}).then(stageObj => {
                stageObj.name = data.filename;
```
This naming of the stageObj. is then the basis for the method `loadStructure` with its `getComponentsByName` function

Maybe this snippet makes the whole process more clear:
https://github.com/arose/ngl/blob/e8713514cfa7100a8313b3ff34a3c94b4125ced4/examples/scripts/component/getByName.js

The idea behind all this as soon as someone picks a molecule it is loaded and saved in the stageObj.  which isn't cleared as long the user does not reload the page. So if the user selects some different molecules but from time to time comes back to its very first selection. it's structure is still loaded in the application so there is no need for another backend call.

This is really important for our specific use case since we want the webserver calls as low as possible and so far it did not lead to too high ram usage of the browser. However I am not sure if that approach would be suitable for mobile users as well.

Up to now we did not any stress tests yet. Meaning how much structures I can load into the stage till this system breaks would be interesting to get to know the limitations. :dagger:

src/lib/components/DashNgl.react.js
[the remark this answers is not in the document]
IvoLeist: The problem is not that it is overwritten but rather it does not compile correctly any longer or especially does not pass your JavaScript lint test. Somehow the render function does literally not see the variable. I guess I need to somehow "import" the variable into the render function but so far I have not figured how...

So weird that it in my component it just works (?)
But to be honest I did not do any JavaScript linting and such a like.

src/lib/components/DashNgl.react.js
IvoLeist: Thanks a lot ! for having a look into I integrated it and it works flawlessly :)

IvoLeist:
@shammamah  The before failed test (pyhton 3.7) magically resolved it self?! 
Glad to see that :) I guess "deploy_to_playground" is supposed to fail ?
Or where do I get the missing `DASH_PLAYGROUND_ENTERPRISE_URL` ?

shammamah:
@IvoLeist Yeah, sometimes we run into timeout errors -- rerunning the test usually resolves it. I'll look into `deploy_to_playground`.

IvoLeist:
@shammamah I did not rerun it. I thought it is only running if you push a commit?
Or is there a way to trigger a rerun manually?

shammamah:
Yes, everyone with admin access to the repository can rerun jobs!

IvoLeist:
@shammamah but I do not yet count as admin right ? So if I want to rerun I have to push a commit ?!

shammamah:
That's correct -- or ping someone who is an admin to rerun the test for you :)

IvoLeist:
@shammamah 
This branch is out-of-date with the base branch
Merge the latest changes from master into this branch.
This merge commit will be associated with Ivo.leist@googlemail.com.

Should I just click update branch ?

shammamah:
Yes please!

shammamah:
OK, so my suspicion is that the CI deployment failure has to do with the fact that this is a fork and not a branch (cc @josegonzalez). Not a problem, I can do a manual deployment for app QA. In the meantime I can also look over the code again.

src/lib/components/DashNgl.react.js
```diff
@@ -0,0 +1,228 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import {Stage, Selection} from 'ngl';
+
+/**
+ * The Molecule3dViewer is used to render schematic diagrams
```
shammamah:
```suggestion
 * The NglViewer is used to render schematic diagrams
```

Could you also please rename the component to just `Ngl` or `NglMoleculeViewer` (or something else along those lines)? That would be more in line with the rest of our components.

IvoLeist:
@shammamah is there any way for me to see the screenshots done by `dash_duo.percy_snapshot`?

shammamah:
@IvoLeist Unfortunately not; only members of the Dash team will be able to do that.

IvoLeist:
@shammamah mh and locally (?) I am about to write a test for changing the background
but the headless browser opens and closes so fast...

Can we maybe extend the the simple_app_callback with the parameter
`wait_time` which sets a wait time (using time.sleep) for just before the percy snapshot
Like I did it in my modification of the simple callback (see test_dashNgl.py)
& integrate something that the screenshots are saved in the test folder ?

Or is this to tedious because then we would have to change a lot of test files ?

shammamah:
What I've done in the past is manually added in a `time.sleep` as I am still writing the test locally. I don't think it makes sense to add a `wait_time` parameter, since it will really only be used locally. If your component just takes some time to render, you can look at https://github.com/plotly/dash-bio/blob/120b1d342d2a2fb61a751a1594af8f32c5faec12/tests/integration/test_ideogram.py#L35 for how to use the `WebDriverWait`.

tests/integration/test_dashNgl.py
```diff
@@ -0,0 +1,215 @@
...
+        "float": "left",
+        "marginTop": "50px",
+        "marginRight": "50px",
+    },
+)
+
+
+# Based on simple_app_callback
+# simple_app_callback does not work because:
+# 1.DashNgl does not accept the output as a string
```
shammamah: Could you elaborate on this? Does it require a dict? If so, you can stringify/unstringify like we've done here: https://github.com/plotly/dash-bio/blob/447ebbe01d716a0d652e744d8fe44b781fa66ad4/tests/integration/test_circos.py#L40

tests/integration/test_dashNgl.py
```diff
@@ -0,0 +1,215 @@
...
+# 2.Two submissions are needed to pass shouldComponentUpdate in DashNgl
```
shammamah: Could you elaborate on this as well?

src/lib/components/DashNgl.react.js
IvoLeist: Ngl Viewer is already take by the developers of Ngl
http://nglviewer.org/

I am not sure if I am allowed to do that since this might lead to confusion between this Dash component and their Jupyter Widget maybe NglMoleculeViewer is better ?

tests/integration/test_dashNgl.py
IvoLeist: yeah but this runs for "ages" because then it stringifies the whole pdb file and paste it in the input box

shammamah: Could you use a smaller file? e.g., https://github.com/plotly/dash-bio/blob/28d37e06d89d271297c526191233419b4893c034/tests/integration/test_molecule3d.py#L16

tests/integration/test_dashNgl.py
IvoLeist: When the app loads an empty dictionary is  loaded with the selectedValue:placeholder
see the console.log screenshots of my app "That is a weird bug if you console.log (this.props.data) during the start of the app its properties are changing. See in the last line in the picture below"

So the thing which has to be passed is:
```
const oldSelection = this.props.data[0].selectedValue;
const newSelection = nextProps.data[0].selectedValue;
if (oldSelection !== newSelection) {
            return true;
        }
```
It is not possible to initialise the container correctly w/o any
dummy value at least I did not manage it

tests/integration/test_dashNgl.py
IvoLeist: I could use a smaller one for the first test but I need for the following tests
a file which has multiple chains and the cif files have even more text.
& for testing several molecules next to each other I need even more data
so stringifiy everything is not really an option.

Especially for the latter one I even would be forced to clear the box in between.
So this work around was the best I could hope for.

But I can assure you I tried for hours to make it work with your callbacks ;D

IvoLeist:
Re 

> What I've done in the past is manually added in a time.sleep as I am still writing the test locally. I don't think it makes sense to add a wait_time parameter, since it will really only be used locally. If your component just takes some time to render, you can look at

It is not about the rendering time but the orientation animations.
But I guess I can set a variable in DashNgl which can be passed from Dash
to the frontend which blocks this behaviour for testing purposes

src/lib/components/DashNgl.react.js
shammamah: Yes, NglMoleculeViewer sounds good to me!

shammamah: The component should only be used in the Dash Bio namespace regardless, so I don't imagine there will be much confusion either way - but better to be safe than sorry :)

tests/integration/test_dashNgl.py
shammamah: OK I see! Yes, please rewrite the first test with a smaller file and keep the other tests the same.

IvoLeist: Ok I can give it a try!

src/lib/components/DashNgl.react.js
IvoLeist: Ok so does this involve renaming all the filenames as well or only inside DashNgl.react.js

tests/integration/test_dashNgl.py
shammamah: I think it might be because you accept both a dictionary/object and a list/array for the `data` prop: https://github.com/IvoLeist/dash-bio/blob/e47ceba4057d00138bec2bf7febb678c2d539bb2/src/lib/components/DashNgl.react.js#L213

You could define a default value as an empty dict or an empty list here: https://github.com/IvoLeist/dash-bio/blob/e47ceba4057d00138bec2bf7febb678c2d539bb2/src/lib/components/DashNgl.react.js#L180

tests/integration/test_dashNgl.py
IvoLeist: I tried it with your file. However, even this file has around 1000 lines. Which means quite some wait time. Haven't found the time yet to let it run completely
but maybe if you have some spare computing resources you can give it a go
see code below:
```
 # Placeholder which is loaded if no molecule is selected
data_dict = {
       "selectedValue": "placeholder",
       "chain": "ALL",
       "color": "#e41a1c",
       "filename": "placeholder",
       "ext": "",
       "config": {"type": "text/plain", "input": ""},
   }

viewer = html.Div(
    [dash_bio.NglMoleculeViewer(
        id=_COMPONENT_ID,
        data=data_dict)],
    style={
        "display": "inline-block",
        "width": "calc(100% - 500px)",
        "float": "left",
        "marginTop": "50px",
        "marginRight": "50px",
    },
)

def test_dbdn_003_show_oneMolecule_pdb(dash_duo):
   data_path =  'tests/dashbio_demos/dash-ngl-moleculeviewer/data/1BNA.pdb'

   fname = '1BNA.pdb'
   d = data_dict.copy() #shallow copy
   d['fname'] = fname
   d['selectedValue'] = fname.split('.')[0]
   d['ext'] = fname.split('.')[1]

   with open(data_path+fname, "r") as f:
      d['config']['input'] = f.read()

   app = dash.Dash(__name__)

   app.layout = html.Div(simple_app_layout(
        viewer,
    ))

   simple_app_callback(
        app,
        dash_duo,
        component_id=_COMPONENT_ID,
        test_prop_name='data',
        test_prop_value=json.dumps(d),
        prop_value_type='dict'
    )
```

josegonzalez:
Yeah @shammamah we disallow secrets from being sent to fork PRs as folks could print those secrets out...

tests/integration/test_dashNgl.py
IvoLeist: @shammamah adding a default dictionary was a really good idea it resolved that issue :)
& this: 

> When the app loads an empty dictionary is loaded with the selectedValue:placeholder
> see the console.log screenshots of my app "That is a weird bug if you console.log (this.props.data) during the start of the app its properties are changing. See in the last line in the picture below"

as well :) !

src/lib/components/DashNgl.react.js
IvoLeist: I renamed all the files

src/lib/components/DashNgl.react.js
IvoLeist: resolved by adding a default dictionary :)

src/lib/components/DashNgl.react.js
shammamah: Please just rename the `DashNgl.react.js` file to `NglMoleculeViewer.react.js`, and re-run the build script - it should auto-update everything :)

IvoLeist: Done :)

IvoLeist:
@shammamah tests are again failing on python-3.7 can you have a look into it?

shammamah:
@IvoLeist Could you try merging `master` into your branch first? We've been seeing this error intermittently, so doing another run might resolve it.

tests/dashbio_demos/dash-ngl/app.py
IvoLeist: @shammamah unfortunately the bug did not vanish completely the mount loop which leads to crashing the app is gone. However,  I realized that my feature: enabling reload from browser when molecule was selected before is not working since the component gets remounted after every new selection.

If you want to test on your own I updated my dash_ngl component
just uncomment dcc.Loading and you will see the component did mount
console.logs appearing after each new selection :-/

IvoLeist:
@shammamah thank you merging it with ` master` indeed resolved the issue :)

IvoLeist:
@shammamah @rpkyle I guess with Corona you have more urgent tasks then 
dealing with my pull request but could you review my current version of the component 
to give me feedback what else is missing before the final merge with the master branch ?

On my wish list there would be:
- [ ] Upload of a structure
- [ ] Support of gzipped files
- [ ] Color changes of a structure

& I might need some support for a javascript test is there any way to test 
```
 // If user has selected structure already used before load it from the browser
 112     loadStructure(stage, filename, chain, color, xOffset) {
```

shammamah:
Hi @IvoLeist - apologies for the radio silence! I'll take another look right now.

src/lib/components/NglMoleculeViewer.react.js
```diff
@@ -0,0 +1,274 @@
...
+        const stage = new Stage(id, params);
+        stage.setSize(viewportStyle.width, viewportStyle.height);
+        this.setState({stage});
+    }
+
+    shouldComponentUpdate(prevProps, nextProps) {
+        const {stageParameters, data} = this.props;
+
+        // check if data has changed
+        if ((data !== null) & (prevProps.data !== null)) {
```
shammamah:
```suggestion
        if ((data !== null) && (prevProps.data !== null)) {
```
A single ampersand is a bitwise operation for two numbers (https://stackoverflow.com/questions/7310109/whats-the-difference-between-and-in-javascript). I think you meant to use `&&` which is the usual Boolean operator.

src/lib/components/NglMoleculeViewer.react.js
```diff
@@ -0,0 +1,274 @@
...
+                return true;
+            }
+        }
+
+        // check for stage params changed
+        const oldStage = prevProps.stageParameters;
+        const newStage = stageParameters;
+
+        // save it as a helper function in an extra script?
+        const isEqual = (obj1, obj2) => {
```
shammamah: You can use Ramda here, like I have done in the `fornaContainer` component: https://github.com/plotly/dash-bio/blob/c920884769f3a3b95d0836f3d5b4216258e0e5a6/src/lib/components/FornaContainer.react.js#L69

src/lib/components/NglMoleculeViewer.react.js
```diff
@@ -0,0 +1,274 @@
...
+
+        // no update since neither the data nor the stage paramas have changed
+        return false;
+    }
+
+    componentDidUpdate() {
+        const {data, stageParameters} = this.props;
+        const {stage, structuresList} = this.state;
+
+        stage.setParameters(stageParameters);
```
shammamah: `componentDidUpdate` is called after the component renders -- why do we need to set the stage parameters here again?

IvoLeist:
@shammamah no worries ;), give me one minute before you dive in the code too deeply I have some pending changes on the uploadStructure branch of my component (https://github.com/IvoLeist/dash_ngl/tree/uploadStructure) which I am about to merge into my dash-bio fork

src/lib/components/NglMoleculeViewer.react.js
```diff
@@ -0,0 +1,274 @@
...
+     * . indicates that only one chain should be shown
+     *  _ indicates that more than one protein should be shown
+     */
+    pdbString: PropTypes.string,
+
+    /**
+     * The data (in JSON format) that will be used to display the molecule
+     * selectedValue: pdbString
+     * color: color in hex format
+     * filename: name of the used pdb/cif file
```
shammamah: Is this a URL or a local app path? Is it possible to pass a string that has the pdb info in it?

IvoLeist:
Information to the newest push: after upload of a structure the dropdown cannot be set to the added structure because you cannot have the same input and output
@shammamah do you maybe have a work around for that?

If you want to see it in action clone https://github.com/IvoLeist/dash_ngl/tree/uploadStructure

shammamah:
@IvoLeist Are you trying to have the following sequence of events: user uploads data -> data file gets added to dropdown as an option? I'm confused about where the circular callback is.

src/lib/components/NglMoleculeViewer.react.js
IvoLeist: good catch thank you :)

src/lib/components/NglMoleculeViewer.react.js
IvoLeist: Ah cool, never used it before. Works as expected :+1:

src/lib/components/NglMoleculeViewer.react.js
IvoLeist: This is needed for changing the viewer style (background, render quality etc.)

src/lib/components/NglMoleculeViewer.react.js
IvoLeist: Do you mean filename? This is already a string e.g. "1BNA.pdb "
It is needed for loading the structure again if it was used before

IvoLeist:
Are you trying to have the following sequence of events: user uploads data -> data file gets added to dropdown as an option?
This is working.

What is not working is:
 user uploads data -> data file gets added to dropdown as an option? --> dropdown selection is changed to the molecule which was just uploaded

src/lib/components/NglMoleculeViewer.react.js
```diff
@@ -0,0 +1,286 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import {Stage, Selection} from 'ngl';
+import * as R from 'ramda';
```
Marc-Andre-Rivet: Due to differences in project configurations and ramda versions behaving differently for default import, the preferred import/usage/syntax for Dash core packages is:
```
import { equals } from 'ramda';
```

src/lib/components/NglMoleculeViewer.react.js
```diff
@@ -0,0 +1,286 @@
...
+        //         return false;
+        //     }
+
+        //     for (const objKey of obj1Keys) {
+        //         if (obj1[objKey] !== obj2[objKey]) {
+        //             return false;
+        //         }
+        //     }
+        //     return true;
+        // };
```
Marc-Andre-Rivet: Commented code block can be removed

src/lib/components/NglMoleculeViewer.react.js
```diff
@@ -0,0 +1,286 @@
...
+
+        if (!R.equals(oldStage, newStage) === false) {
+            return true;
+        }
+
+        // if (isEqual(oldStage, newStage) === false) {
+        //     return true;
+        // }
```
Marc-Andre-Rivet: Commented code block can be removed

src/lib/components/NglMoleculeViewer.react.js
```diff
@@ -0,0 +1,286 @@
...
+                        data.filename,
+                    ]),
+                }));
+            });
+    }
+
+    render() {
+        const {id} = this.props;
+        return (
+            <div>
```
Marc-Andre-Rivet: Just wondering why there's a need for a wrapper `div`

src/lib/components/NglMoleculeViewer.react.js
Marc-Andre-Rivet: Seems like `stage` and `structuresList` could be instance variables or the result of a memoized/calculated function call that would get mutated during `render` instead of using `didUpdate` / `didMount` to mutate them (which goes against React's paradigm: states/props are immutable)

Due to the experimental nature of this component addition, I'm fine with the way this is implemented for now but we should have plan to follow up / clean up later as required.

src/lib/components/NglMoleculeViewer.react.js
IvoLeist: forgot to delete it thank you :)

src/lib/components/NglMoleculeViewer.react.js
IvoLeist: good catch removed the unnecessary wrapper

src/lib/components/NglMoleculeViewer.react.js
IvoLeist: @Marc-Andre-Rivet I was not aware that this is against  Reacts paradigm. Thank you for pointing that out. Do you know some good resources where I can learn about the correct usage of instance variables/memoized/calculated functions ?

IvoLeist:
@shammamah  can you try if a rerun fixes the errors?
Results (360.93s):
      61 passed
       2 failed
         - tests/integration/test_forna_container.py:98 test_dbfc003_color_scheme_preset_structure
         - tests/integration/test_forna_container.py:139 test_dbfc004_color_scheme_preset_position

The newest push ticks all the items on my whishlist
- [x] Upload of a structure
- [x] Support of gzipped files
- [x] Color changes of a structure

For working prototype clone:
https://github.com/IvoLeist/dash_ngl

What features are on your whishlists?

Marc-Andre-Rivet:
@IvoLeist Sorry for the long delays over the last few rounds. I'll be reviewing the PR from this point on. Re-ran the CI jobs, there's apparently some flakiness but they passed eventually. For `deploy_to_playground` since it requires specific knowledge of our deployment environment, we'll need to see if anything breaks once it's merged into master and fix on our side accordingly. Same goes for making the component async.

I see one scenario that's did not work: https://github.com/plotly/dash-bio/pull/496#issuecomment-603435718, https://github.com/plotly/dash-bio/pull/496#issuecomment-604003500, can you provide more details? Does this work now? I don't think I see anything else to wrap up apart from manual testing of the component on my side.

IvoLeist:
@Marc-Andre-Rivet  no worries in Corona times everything is a little bit slow.
Cool that you guys are willing to make the component async.

Have not found the time yet to figure out a workaround for https://github.com/plotly/dash-bio/pull/496#issuecomment-603435718 but regarding deployment this should not hinder you since it is just a "nice to have" feature.

Some information to the newest commits:
- added the functionality to specify the amino acid range of the chosen molecule's chain
- added the functionality to save the output of the viewer as png

Regarding manual testing how do you do that. When I want to run
any app in tests/dashbio_demos/ it always runs in:
`ModuleNotFoundError: No module named 'layout_helper'`
Therefore I have that [repo](https://github.com/IvoLeist/dash_ngl) for my manual testing

Off topic: 
Do you know if the [dash-cytoscape](https://github.com/plotly/dash-cytoscape) repo is still maintained? Because the web app we are developing right now does not only need that dash-ngl component but also some extra functionalities for dash-cytoscape. Meaning accessing the image generation API of cytoscape.js. Like it is done in that [PR](https://github.com/plotly/dash-cytoscape/pull/65)

IvoLeist:
Information to the latest [commit](https://github.com/plotly/dash-bio/pull/496/commits/a26509ab1c17b353b36446d75155a7544a819c17)
Added the possibility to change the molecule representation:
* Possible molecule styles:
     'backbone,'ball+stick','cartoon', 'hyperball'
     'licorice','line','ribbon','rope','spacefill',
     'surface','trace','tube'

* Possible additional representations:
     'axes','axes+box','helixorient','unitcell'

IvoLeist:
Information to https://github.com/plotly/dash-bio/pull/496/commits/3377720bc17a8e299000ad9f17d4fa692620339f 

<p align="center">
  
</p>

yarn.lock
```diff
@@ -0,0 +1,12889 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Marc-Andre-Rivet, line 1: A bit confused by this addition. We don't use `yarn` in this project.

IvoLeist: sorry for the confusion I just played around with yarn because some one asked about it https://github.com/plotly/dash-bio/issues/502
I removed it see newest commit

IvoLeist:
@Marc-Andre-Rivet any idea why it is failing:
This is what I found in the logs:

> ERROR: requests 2.21.0 has requirement urllib3<1.25,>=1.21.1, but you'll have urllib3 1.25.9 which is incompatible.

> dash-renderer@1.4.1 prepublishOnly /home/circleci/project/dash-main/dash-renderer
> rm -rf lib && babel src --out-dir lib --copy-files

> sh: 1: babel: not found
> npm ERR! code ELIFECYCLE
> npm ERR! syscall spawn
> npm ERR! file sh
> npm ERR! errno ENOENT
> npm ERR! dash-renderer@1.4.1 prepublishOnly: `rm -rf lib && babel src --out-dir lib --copy-files`
> npm ERR! spawn ENOENT
> npm ERR! 
> npm ERR! Failed at the dash-renderer@1.4.1 prepublishOnly script.
> npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
> npm WARN Local package.json exists, but node_modules missing, did you mean to install?
> 
> npm ERR! A complete log of this run can be found in:
> npm ERR!     /home/circleci/.npm/_logs/2020-05-07T14_00_47_156Z-debug.log
> npm ERR! code ELIFECYCLE
> npm ERR! errno 1
> npm ERR! dash-renderer@1.4.1 build: `npm run prepublishOnly && renderer build`
> npm ERR! Exit status 1
> npm ERR! 
> npm ERR! Failed at the dash-renderer@1.4.1 build script.
> npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
> npm WARN Local package.json exists, but node_modules missing, did you mean to install?
> 
> npm ERR! A complete log of this run can be found in:
> npm ERR!     /home/circleci/.npm/_logs/2020-05-07T14_00_47_166Z-debug.log
> 
> Exited with code exit status 1
>

Marc-Andre-Rivet:
@IvoLeist Yes. A change in the renderer requires a change in dash-bio's `config.yml`

https://github.com/plotly/dash-bio/blob/master/.circleci/config.yml#L37
`cd dash-renderer && npm run build && pip install -e . && cd ../..`
needs to change to:
`cd dash-renderer && npm ci && npm run build && pip install -e . && cd ../..

IvoLeist:
@Marc-Andre-Rivet looks like we committed the same change :smile: 
Anyway python3.6 test is passed now. Python3.7 runs in some issues, see below:

> Results (384.93s):
>       67 passed
>        1 failed
>          - tests/integration/test_ideogram.py:13 test_dbid001_displayed_chromosomes
> 
> Exited with code exit status 1

Marc-Andre-Rivet:
@IvoLeist The py37 test failure seems to have been only flakiness. Let me know if there's more you plan on adding to this PR / component. On my side it seems about ready to go. `deploy_to_playground` can't work for obvious reasons so we'll have to test it out once it makes it in. I'll follow up by making the component async afterwards. For now it will be possible to use/install this version of the component directly from GitHub but as the last two points above get done / tested, we'll make a follow up release of it.

IvoLeist:
@Marc-Andre-Rivet fingers crossed that the tests of the newest commit run through smoothly.
Just added the possibility to set a default filename for a saved structure image and the enabled changing the molecules spacing from Python. There is always something which might be needed in the future but for now I would say I am ready to go as well :rocket:

IvoLeist:
@Marc-Andre-Rivet 
the test 13 in ideogram is again "flaky" (?) but this time it happened in python-3.6
> Results (449.03s):
>       69 passed
>        1 failed
>          - tests/integration/test_ideogram.py:13 test_dbid001_displayed_chromosomes

Some information to the latest push:
We realised in our app that once an image downloaded is selected the download handler gets triggered constantly this push fixes it. On top I added an integration test, which tests the download functionality

If you wonder why I added `dash-main/` to the .gitignore: 
I had to update my venv with the requirements.txt from [dash-main](https://github.com/plotly/dash) because the test rotate structure and reset it (dbn_16) was failing before. No clue which of the building dependencies caused that issue...
anyway it is now working :)

IvoLeist:
@Marc-Andre-Rivet, @rpkyle, @shammamah, @josegonzalez  sorry for pinging you all 
but I was not sure where/whom to ask that question. 

We recently deployed this component as a part of our new app. However, for some of our test users everything worked fine while for others the component run in plenty of errors. Tracked them down I realized that they all were related to webGL being either disabled or their GPU was on the black list. Thus, setting the correct flags in chrome (Brave) or changing the about:config in firefox fixes that bug.

The current work around is a big disclaimer box explaining the user when there is no structure appearing how they do have to change the settings of their browser. Obviously, I only want to show that box if the viewer runs in a webGL error, so do you have a good idea how to catch that error?

Do you maybe have some kind of best practice for your dash-bio/dash/plotly apps for catching that?

alexcjohnson:
I'm not sure if we explicitly handle the no webgl case elsewhere in dash-bio, but in plotly.js we do - I think there are a number of things it looks for, but at first cut you could just wrap the first call that typically errors in a `try/catch`, and bail out after that.

Here's where plotly.js does this for its 3D plots: https://github.com/plotly/plotly.js/blob/be708650801e24a0c9eb56bd6bdc755834cb8daa/src/plots/gl3d/scene.js#L194-L201

jackparmer:
@IvoLeist Awesome to hear that you got this component working in your app! I'd love to see screenshots if you can share. 

Also curious if you would like to wrap up this PR, or if you have any guidance for our team if we do.

I'm interested in making a Dash app that plots these `pdb` files and compares the 3 different molecule renderers in Dash Bio (dash-ngl, dash-speck, and dash-mol3D):

https://deepmind.com/research/open-source/computational-predictions-of-protein-structures-associated-with-COVID-19

IvoLeist:
@alexcjohnson @Marc-Andre-Rivet @shammamah  @rpkyle 
Hey guys,

I am sorry that I was suddenly so silent but I was quite "spontaneously" switching my working place (CNAG) and my place to live (Barcelona 😎). In order to have a good start in my PhD I had to catch up as quickly as possible with the tools they are using there. Since I like dash I'll try to integrate in the web app I am building there but I cannot promise anything because their apps are at the moment React only but developing that component did definitely help in getting into React.

Short story short I guess I unfortunately won't find too much time to wrap up that PR.
Doing only some bugfixes for my old employer.
But since It works kinda flawlessly in the app of my old employee it should not be too difficult to polish
to finally merge it ;D. I let you know as soon as it is published.
Maybe he finds the time to finish that PR?

@jackparmer cool to hear that you show interest if it is only a short rendering of the molecule.
I guess my toy app in my repo should work out of the box https://github.com/IvoLeist/dash_ngl
If you have some troubles to let it run just file an issue there so we do not Spam too much into that PR.
Btw. this is actually a cool idea to show the power of my component. Should have done this anyway. 
So I would appreciate if you fork it and show me the results would be really interested.

IvoLeist:
Ok I was just too intrigued  by it  and hat too see the structures myself.
@jackparmer I made a picture for you so you understand better what the special power of my component is.
You see on it how easy it is too show several protein structures (works only if you select a chain)
side by side. This was not possible with any of the existing viewers. That's why I have built it on my own.
For the general public this feature might be too niche and thus the whole component maybe too tailored to our specific use case.
dash-ngl_covid_SideBySide


But you can also use it as normal viewer. Here the structures of NP4,NP6,NP8 are merged into one.
Unfortunately they are not colored differently since this only works for chains but I guess this should not be too difficult to add that.
dash-ngl_covid_merged
What I would consider as a USP in comparison too the other viewers out there that mine is checking if a structure was already loaded into the frontend and if so it does not load it again from the backend but directly from the frontend.

jackparmer:
Wow, this side-by-side view is beautiful! I was a bit puzzled when I first saw this PR, but I see the advantage now over the Speck and Mol3d renderers.
Is it possible to drag the structures together to see how molecules intercalate (like ChemDraw)?
Do you mind pushing this app to GitHub somewhere? I'd love to play with it. 
With your permission, I'd also like to try deploying it and sharing it with the DeepMind team (I wonder if this component could be useful in a Dash tool for AlphaFold research).

IvoLeist:
@jackparmer 
Glad that you like it was actually quite some effort to get the orientation and the distance right for a nice side by side view.
I just tried it and it seems to be that I at one point deactivated the dragging or I forgot the shortcut combination to make it draggable 😄

I just pushed the newest version (including NSP2,NSP4 etc.) to my repo:
https://github.com/IvoLeist/dash_ngl
Unfortunately right now only NSP2,4,6 are working since it at the moment only supports structures with the 4 letter pdb code.

For sure - actually it would be awesome! Would love to contribute in any way fighting covid / helping research in general.
Let me know if you need some extra features (like the dragging) and I can see how 
I can support you with that

jackparmer:
Not sure if it's possible with ngl, but independent molecule dragging on the same canvas would be awesome. 

Drag is possible with the Mol3d and Speck renderer (magic key for drag with Speck is Shift and Ctl for Mol3d), but I do not believe that you can independently drag multiple molecules on the same canvas with these renderers (is that correct @justinmc @wwwtyro ? ). 

If this worked, I'd love to make some Chem3d-style intercalation demos with the ability to rotate + drag several molecules on the same canvas.

wwwtyro:
@jackparmer Yeah, that's correct for Speck. :+1:

IvoLeist:
@jackparmer 
In theory it should be possible see for example that code pen 
https://codepen.io/xrobin/pen/wXmGmm

Here using
drag-ctrl+shift + right click = pans 
drag-ctrl+shift + left click = rotates 
the cylinder and structure independently
but I cannot really explain why this functionality seems to be disabled in my implementation.

In my code pen they are also not independently dragable ... 
https://codepen.io/IvoLeist/pen/WNQEGML
on a quick glance I was not able to spot the difference between this two but if you are able to modify my pen in such a way that the structures are dragable I guess it should not be difficult to change the dash-ngl component accordingly :v:

xhlulu:
@IvoLeist That's really cool! I'm trying it out with DM's AlphaFold predictions, and I'm happy to share the prelim results once it's more presentable.

I'm wondering: If I have a file named `M_protein.pdb` and I want to display it along my own `nsp2.pdb`, what's the best wait to rename it? I'm presuming that I could rename it `Mpro.pdb` and add the full name in the dropdown? But i'm worried it might change the meaning here.

justinmc:
@jackparmer True for my unfinished library tinymol, unsure for Mol3d.

jackparmer:
Hey @IvoLeist , Good idea to use a CodePen. I played with it and concluded as you did - Ctl-drag only pans the canvas as a whole.

However, unlike the Speck and Mol3d ([molecule-3d-for-react](https://github.com/Autodesk/molecule-3d-for-react)) components, it does seem like there is a way to adjust the relative position of multiple proteins on the same canvas with ngl. This could be done by exposing the Position, Rotation, and Scale parameters as props in the ngl React component (so they would then be optional properties of the ngl Dash component).

image

I may be overlooking something, but I didn't see these parameters in [NglMoleculeViewer.py](https://github.com/plotly/dash-bio/pull/496/files#diff-e72f81097ea42f2f28403b70d9437a70)

If Position, Rotation, and Scale were available as Dash component properties, you wouldn't be able to click-drag proteins independently, but a Dash user could create buttons or use the arrow keys to fire Dash callbacks that adjust the relative positions of multiple proteins. 

I also played with the "official" [ngl JS demo](http://nglviewer.org/ngl/) and had good success changing Position, Rotation, and Scale with 3 of the AlphaFold COVID proteins on the same canvas:

image

IvoLeist:
@xhlulu that is interesting using my upload your M_protein.pdb gets ubploaded correctly but not the whole filename get inserted in the dropdown but just protein. Looking at the code it is pretty obvious why this happens:

This part (shown below) gets the pdb_id directly from the file.
Maybe this is not a a good idea if you want users to allow custom made pdb files.
```
pdb_id = content.split('\n')[0].split()[-1]
        if 'data_' in pdb_id:
            pdb_id = pdb_id.split('_')[1]
            ext = 'cif'
```
Have not tried yet what is happening if you hardcode it, will try it soon.

IvoLeist:
@jackparmer This is definitely possible to expose these properties to the dash component I actually did it for the x axis spacing between each molecule chain. So the user can decide on its own how much the chains should be separated from each other but doing everything with Buttons feels to hacky for me. But maybe it is still a good idea to expose that settings if user want to hardcode the positioning of several molecules (?)

I would rather suggest to look deeper into my JS coding to see what is different to the first code pen and mine.
Since it works flawlessly in the one of xrobin (https://codepen.io/xrobin/pen/wXmGmm)

Since my time is right now quite limited I have asked the colleague in which app that component is working flawlessly.
Maybe he can help :crossed_fingers:

IvoLeist:
@xhlulu 
You were right renaming is possible but only if you hardcode it into the list.
I just updated my example app accordingly:
https://github.com/IvoLeist/dash_ngl/commit/a1bdd83cb7993708a040c33adc88c9d192904935

So if you pull it now and put in the show multiple structures you see that two structures.
Make sure that you select for each a chain e.g. MPRO.A_NSP2.A

What I learned from this:
Using an underscore to separate strings was maybe a bad idea...
I wanted to use something which cannot be exploited for code injection so 
I refused to use $, | etc. but maybe an equal sign would be better :thinking: 

To give you some context why my example is like that.
In the actual app the combination of molecule chains is generated on the fly.
Thus, we needed some clever way to specify which molecules/parts have to be used. 
And the best solution we came up with was to just sent a string from the backend to the frontend.
So for testing this example app was perfect but for actual usage as a Mol3D replacement (?)
I guess it would be better to use the [dash multi value dropdown](https://dash.plotly.com/dash-core-components/dropdown).

The only issue I have with that how would I "specify a chain/ residues range/ highlight chosen residues/ atoms"?
Like it is done now: e.g. 6CHG.A:1-450@a50,a100,a150_3K8P.D
Anyone has an idea? :thinking:

IvoLeist:
Used the "soft" Corona Lockdown to figure out what the difference between 
[that codepen](https://codepen.io/xrobin/pen/wXmGmm) -> structures can be moved independently 
[my codepen](https://codepen.io/IvoLeist/pen/WNQEGML) --> structures can be moved only together

My finding are summarized in [this codepen](https://codepen.io/IvoLeist/pen/vYGOBzb)

```
//do not add to stage --> mols can be moved indepently by
// drag-ctrl+shift + right click
this.loadStructure("1blu","A","red",false)
this.loadStructure("1crn","A","blue",false)

//add to stage --> mols can only moved together
// this.loadStructure("1blu","A","red",true)
// this.loadStructure("1crn","A","blue",true)
```

Why do I need to add the structures in my component to the stage?:
 -> In order to do the autorotation using principal components and show the proteins side by side
See [the codepen](https://codepen.io/IvoLeist/pen/WNQEGML)

Implications for the PR:
To add the desired feature to move/rotate the proteins individual w/o
breaking the app which already uses the component.
I would suggest to add an extra button (variable in the React Component) which gives the user the option to decide if
he/she wants the molecules shown side by side.
Indicating that the side by side view is experimental and thus individual protein movements/rotations by mouse using
drag-ctrl+shift + left/right click do not work.

IvoLeist:
@jackparmer 

Just merged a [quick workaround](https://github.com/IvoLeist/dash_ngl/commit/a774b052d67f03fc665e1e969d5d4f61f85ceb85) into my master 

The new radio selection is connected to the react component in that way that it ether adds
the structure to the stage (for the side by side view) or not (for the normal view).
In the latter case the molecules can be dragged/ rotated individually by ctrl and left/right click.
! Note: it only rotates if you click on the molecule not in the box
you see a mouse over if you have the molecule selected correctly.

Let me know if it works for you as well :crossed_fingers: 
Before adding this workaround to the PR I would wait for the answer of the guys from ngl
since this hack does not feel "right". Anyway I am happy that it at least works in the not side by side view :)

dash-ngl_covid_moveIndependently

jackparmer:
Wow, this component just gets better and better. Are you connected with the
NGL core devs?

IvoLeist:
Glad that you like it. Unfortunately not directly
So far I only asked via filing a github [issue](https://github.com/nglviewer/ngl/issues/777)

They merged a PR 4 days ago so it still looks maintained.

But if you have some contacts to the NGL core devs 
feel free to use them it might speed up things a little :crossed_fingers:

xhlulu:
> @jackparmer
> 
> Just merged a [quick workaround](https://github.com/IvoLeist/dash_ngl/commit/a774b052d67f03fc665e1e969d5d4f61f85ceb85) into my master
> 
> The new radio selection is connected to the react component in that way that it ether adds
> the structure to the stage (for the side by side view) or not (for the normal view).
> In the latter case the molecules can be dragged/ rotated individually by ctrl and left/right click.
> ! Note: it only rotates if you click on the molecule not in the box
> you see a mouse over if you have the molecule selected correctly.
> 
> Let me know if it works for you as well 
> Before adding this workaround to the PR I would wait for the answer of the guys from ngl
> since this hack does not feel "right". Anyway I am happy that it at least works in the not side by side view :)
> 
> dash-ngl_covid_moveIndependently

@IvoLeist Reallly cool! I just cloned your repo and I'm trying to reproduce it locally, but I can't get the image you are showing. I can modify the textbox to be `nsp2.A_nsp4.A_nsp6.A` to show them side by side, but then I can't drag them separately with ctrl+click.

Could you make a short GIF showing how you made it work?

IvoLeist:
@xhlulu sorry for the delayed follow up - I was on vacations.

Cool that you like it, but very unfortunate that it is not working for you :-/
Did you figured it out in the mean while?

@jackparmer regarding replacing my workaround:
Got some insight from the devs of ngl https://github.com/nglviewer/ngl/issues/777
like I already expected the issue is not on my side but they have to "patch" their viewer.
Since I am quite busy at the moment I am afraid but I cannot figure out the math.

But if one of your devs has the necessary math skills and some capacity left:
> if you figure out what the relevant transformations are you could even add it to an event listener (listen to component.signals.matrixChanged on the parent (non-displayed molecule), calculate the translation required and apply it to the second component (the one with the representations).

xhlulu:
> @xhlulu sorry for the delayed follow up - I was on vacations.
> 
> Cool that you like it, but very unfortunate that it is not working for you :-/
> Did you figured it out in the mean while?
> If not here is video showing how to drag/rotate proteins individually.
> 
> https://github.com/IvoLeist/dash_ngl/blob/master/static/demo_4.mp4
> !Note this is only working if you have selected show molecules side by side?: No

Thanks for sharing this! I didn't know about the right click - i'll look into it when i get some time and let you know if the right click works for me!

jackparmer:
Some nice clues in https://github.com/nglviewer/ngl/issues/777 - Thanks for investigating @IvoLeist !

I believe @HammadTheOne is going to take over this repo's open PRs next month and try to bring them to the finish line this fall. Thanks for all of the initial work and consultation. We'll get this in!

IvoLeist:
@jackparmer that are awesome news :rocket: 
I try to watch this PR from time to time to make sure that it in the end is still compatible to the app in which it is already running :crossed_fingers: 

@HammadTheOne do not hesitate to ping me if you have some issues always happy to help ;D
Thank you for picking up this PR - all the best!

CONTRIBUTING.md
```diff
@@ -154,12 +154,14 @@ To run integration tests locally on, say, Google Chrome:
 python -m venv venv
 . venv/bin/activate
 pip install -r tests/requirements.txt
-pip install dash[testing]
-pytest tests/integration
+pip install dash[testing] #if you use zsh run: pip install dash\[testing\]
+pip install dash-bio #if pytest fails try: pip3 install dash-bio
+pytest tests/integration #for testing all apps
+pythest tests/integration/test_yourNewApp #for testing only one app
```
alexcjohnson, line 160:
```suggestion
pytest tests/integration/test_yourNewApp #for testing only one app 
```

R/dashbioDashNgl.R
```diff
@@ -0,0 +1,18 @@
+# AUTO GENERATED FILE - DO NOT EDIT
+
+DashNglDashNgl <- function(id=NULL, viewportStyle=NULL, stageParameters=NULL, data=NULL) {
```
alexcjohnson, line 3: What is this file? Obsolete?

HammadTheOne: Yes, thanks for pointing it out. I've rebuilt the R package and deleted this file in 9f05326.

src/lib/components/NglMoleculeViewer.react.js
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * The height and the width (in px) of the container
+     * in which the molecules will be displayed.
+     * It should be in JSON format.
+     */
+    viewportStyle: PropTypes.exact({
+        width: PropTypes.string,
+        height: PropTypes.string,
+    }),
```
alexcjohnson, line 93: Can we break this out into two props, `width` and `height`, each of which is just a number? The current implementation expects a CSS string, which is indeed more flexible but it's harder to use for non-web-devs. If you think this flexibility is important we could support both, like `widthStr = isNumeric(width) ? (width + 'px') : width` (using my favorite [`fast-isnumeric`](https://www.npmjs.com/package/fast-isnumeric) 😉 so `500` and `'500'` count as numbers, ie pixels, but any other string is a CSS size string) or we can start with just numbers and include strings later if there's interest.

alexcjohnson:
Apparently the component works without it, but we should add `setProps` as an explicit prop.

src/lib/components/NglMoleculeViewer.react.js
HammadTheOne: Added in 78fd7e9.

src/lib/fragments/NglMoleculeViewer.react.js
```diff
@@ -22,12 +23,19 @@ export default class NglMoleculeViewer extends Component {
...
 
     componentDidMount() {
-        const {id, stageParameters, viewportStyle} = this.props;
+        const {id, stageParameters, viewportWidth, viewportHeight} = this.props;
         const params = {...stageParameters};
         const stage = new Stage(id, params);
         const orientationMatrix = stage.viewerControls.getOrientation();
 
-        stage.setSize(viewportStyle.width, viewportStyle.height);
+        const widthStr = isNaN(viewportWidth)
```
alexcjohnson, line 31: `isNaN` has some weird properties - for example it thinks whitespace strings are numbers (because `+' ' === 0`. So please just use `fast-isnumeric`.

src/lib/components/NglMoleculeViewer.react.js
```diff
@@ -83,14 +83,21 @@ NglMoleculeViewer.propTypes = {
...
-        width: PropTypes.string,
-        height: PropTypes.string,
-    }),
+    viewportWidth: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+
+    /**
+     * The height (in px or as a number) of the container
+     * in which the molecules will be displayed.
+     */
+    viewportHeight: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
alexcjohnson, line 100: Can we just use `width` and `height`? That matches what we do for the overall size of a lot of other components, and doesn't seem ambiguous to me so I don't think we need the `viewport` qualifier.

src/lib/fragments/NglMoleculeViewer.react.js
HammadTheOne: I thought `isNaN` might be a little cleaner so we're importing fewer modules, but `isNumeric` does make more sense and removes ambiguity in this case. Updated in c39b8a6.

src/lib/components/NglMoleculeViewer.react.js
HammadTheOne: Updated in c39b8a6.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [0.4.9] - 2020-10-05
+### Changed
```
alexcjohnson, line 4:
```suggestion
## [0.5.0] - 2020-10-05
### Added
```
Oops, I just noticed this now - this is an addition so we should give it a minor bump. We didn't do that for a few additions earlier in the year, but I think this would be a better pattern.

alexcjohnson (approved):
Looks great! Just one final comment about the version number, then let's go! 💃

PR #497, opened 2020-03-09
graph_objs --> graph_objects

emmanuelle:
This PR proposes to bump the plotly.py dependency to the 4.x series, and modifies accordingly some components with the new syntax `graph_objects` instead of `graph_objs`. (4.0 was released in July 2019)


emmanuelle:
also @nicolaskruchten

Issue #474, opened 2020-01-24
Our Percy snapshots aren't being properly stored. 


shammamah:
**Describe the bug**
There is one Percy diff for a graph (`test-graph_effect_size_line_[-0.5, 1.5] - py3.6`
) that Percy believes is a new snapshot each time we open a PR.

PR #512, opened 2020-08-10
Dendrogram bug

### LINKED ISSUE #507: Clustergram - dendrograms not appearing?
Look at https://dash.plotly.com/dash-bio/clustergram - the dendrograms don't appear, only the heatmap.

cc @nicholas-esterer - related to the work you're doing in plotly.py on dendrograms?

nicholas-esterer:
Closes #507 

## About
- I am closing an issue.

## Description of changes
It looks like the dendrogram part of the clustergram code was written before plotly.py had its own dendrogram figure factory. So the fix here was to scrap the old dendrogram code from this repo and just use plotly.py's implementation (which also has some recent bug fixes related to colormaps). The new colors do not match the ones that used to show up with the dendrogram code coming from this repository and scipy==1.4.1, but the code computing the colors (`_color_dendro_clusters` and its subroutines) is very hard to follow and its purpose is not entirely clear (at least to me). So for that reason I have not tried to preserve the old color scheme, which seemed arbitrary as far as I can tell.


nicholas-esterer:
In order to get the tests to pass it might be necessary to do some plumbing in CI. For example, it looks like pylint is old and suffers from this bug https://github.com/psf/black/issues/48 and it cannot find something called `babel` when installing `dash`.

nicholas-esterer:
Reviewers can run `python examples/clustergram.py` to see the clustergram in action (after setting up a virtual environment and installing dependencies from `requirements.txt`).

dash_bio/component_factory/_clustergram.py
```diff
@@ -384,7 +384,7 @@ def figure(self, computed_traces=None):
         # [row dendro] [heatmap]     [heatmap]     [row GM]
         # [row dendro] [heatmap]     [heatmap]     [row GM]
         # [empty]      [col. GM]     [col. GM]     [empty]
-        fig = tools.make_subplots(
+        fig = subplots.make_subplots(
```
alexcjohnson, line 387: Oh nice, presumably this will remove the warning you get when running dash-docs locally

```
/Users/alex/Library/Python/3.8/lib/python/site-packages/plotly/tools.py:461: DeprecationWarning:

plotly.tools.make_subplots is deprecated, please use plotly.subplots.make_subplots instead
```

Though - you see this warning twice in a row, maybe there's another use of `tools` in `dash-bio`?

alexcjohnson: Nah, looks like only one `make_subplots` anywhere in `dash-bio` - so the docs probably just call this twice.

examples/clustergram.py
```diff
@@ -0,0 +1,61 @@
...
+        options=[
+            {'label': row, 'value': row} for row in list(df.index)
+        ],
+        value=rows[:10],
+        multi=True
+    ),
+
+    html.Div(
+        id='my-clustergram',
+	children=dcc.Graph(figure=clustergram)
```
alexcjohnson, line 55:
```suggestion
        children=dcc.Graph(figure=clustergram)
```
Somehow a hard tab snuck in 😉

examples/clustergram.py
```diff
@@ -0,0 +1,61 @@
...
+app.layout = html.Div([
+    "Rows to display",
+    dcc.Dropdown(
+        id='clustergram-input',
+        options=[
+            {'label': row, 'value': row} for row in list(df.index)
+        ],
+        value=rows[:10],
+        multi=True
+    ),
```
alexcjohnson, line 51: The clustergram in this example looks great! But you're not using this dropdown at all, should either take it out or connect its output to the figure via a callback.

examples/clustergram.py
```diff
@@ -0,0 +1,61 @@
...
+    height=800,
+    width=700,
+    color_list={
+        'row': ['#636EFA', '#00CC96', '#19D3F3'],
+        'col': ['#AB63FA', '#EF553B'],
+        'bg': '#506784'
+    },
+    line_width=2
+)
+clustergram.write_json('/tmp/clustergram.json')
```
alexcjohnson, line 40: for debugging? 🔪 ?

nicholas-esterer: yes I thought I'd rebase-out this before merging :-)

rpkyle:
> Aside from that and my couple of comments on the example app, it's a 💃 from me. I see @rpkyle has some other WIP to fix these tests, I'll let you and him work out the logistics of how, where, and when to get all the tests running again, be it in this PR or another.

@alexcjohnson Yeah, this is great! I think the `Dendrogram` issue is distinct from the fix required for `Clustergram`, so best to merge this PR and fix one set of problems while we tackle another. Hopefully @HammadTheOne and I can sort out the latter without too much fuss.

> In order to get the tests to pass it might be necessary to do some plumbing in CI. For example, it looks like pylint is old and suffers from this bug psf/black#48 and it cannot find something called babel when installing dash.

@nicholas-esterer I fixed the `babel` issue in https://github.com/plotly/dash/pull/1362, but thanks for noting the former issue -- I'll update `pylint` also.

nicholas-esterer:
@rpkyle Great thank you! I'm assuming you will only fix the `pylint` error related to psf/black#48 and not the linting errors regarding `protected-member` and the like. These latter errors I have circumvented with directives in comments and would prefer to keep them around. Also I wouldn't want to disrupt the requirements of the entire project :joy:.

rpkyle:
> @rpkyle Great thank you! I'm assuming you will only fix the `pylint` error related to [psf/black#48](https://github.com/psf/black/issues/48) and not the linting errors regarding `protected-member` and the like. These latter errors I have circumvented with directives in comments and would prefer to keep them around. Also I wouldn't want to disrupt the requirements of the entire project 😂.

😹 Yes, I was just going to bump the `pylint` version to address https://github.com/psf/black/issues/48, and confine my tinkering to that particular issue alone.

rpkyle:
@nicholas-esterer OK, looks like 3.6 and 3.7 tests are passing after all your hard work once I disabled `bad-continuation` in `.pylintrc`. I believe they're actually in the process of removing this check from `pylint` completely, but it hasn't landed in `pip` yet: https://github.com/PyCQA/pylint/issues/3761. 

Once it does, I can update `.pylintrc` to remove the check. I might try to resolve https://github.com/plotly/dash-bio/issues/474 tomorrow so that all tests pass, including Percy.

PR #516, opened 2020-08-16
Igv

shday:
## About
- [ ] I am closing an issue
- [ X] This is a new component
- [ ] I am adding a feature to an existing component, or improving an existing feature

## Description of changes
This is a PR for a new component, Igv (interactive genome visualization). I've used DRAFT mode as the tests currently rely on an internet connection. Need feedback on how best to get around that.

PR #537, opened 2021-01-13
Add igv component

HammadTheOne:
Closes plotly/dash-core#250

## About
- [x] I am closing an issue
- [x] This is a new component

## Description of changes

This PR is a continuation of the draft PR created by @shday [here](https://github.com/plotly/dash-bio/pull/516). This PR includes the following additions to the original PR:

- Updates to `igv.react.js` to enable async loading of the component and greatly reduce bundle size.
- Fixes to enable passing of properties through a Dash app by removing and then re-rendering the IGV browser upon component change. 
- Renaming and fixing of tests to allow them to work with local datasets with `app.get_asset_url` to pass a URL directly to the IGV browser.
- Added a simple demo app to showcase some features and test prop changes.
- Changed `igv` package from experimental `react` fork to (now `react`) compatible `master` fork.


jackparmer:
When you add docs for this, can you please also add docs for dash-ngl at the same time, since it never got documented over at https://dash.plotly.com/dash-bio

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dash-bio",
-  "version": "0.5.0",
+  "version": "0.5.2",
```
alexcjohnson, line 3: Presumably this will get sorted out at publishing time, but to be clear this should be 0.6.0 as it's adding a new component.

src/lib/fragments/Igv.react.js
```diff
@@ -0,0 +1,61 @@
...
+import igv from 'igv';
+
+/**
+ * The Igv component is an interactive genome visualization component
+ * developed by the Integrative Genomics Viewer (IGV) team. It uses an
+ * example integration of igv.js and React (https://www.npmjs.com/package/igv).
+ */
+export default class Igv extends Component {
+    componentDidMount() {
+        var igvContainer = document.getElementById(this.props.id);
```
alexcjohnson, line 12: Don't use `getElementById` - use a `ref` instead. For an example, see https://github.com/plotly/dash-core-components/pull/604/files where this change was made to `dcc.Graph`

src/lib/fragments/Igv.react.js
```diff
@@ -0,0 +1,61 @@
...
+                locus: this.props.locus,
+                reference: this.props.reference,
+                minimumBases: this.props.minimumBases,
+                tracks: this.props.tracks,
+            };
+            igv.createBrowser(this.igvContainer, this.igvOptions).then(function(
+                browser
+            ) {
+                igv.browser = browser;
+            });
```
alexcjohnson, line 49: 🌴 Everything inside this `if` block matches `componentDidMount` (except for setting instance vars `this.igvContainer` and `this.igvOptions` that seems to be useless.) - so let's pull them out into a method that both lifecycle methods call.

src/lib/fragments/Igv.react.js
```diff
@@ -0,0 +1,61 @@
...
+
+    componentDidUpdate(prevProps) {
+        if (
+            this.props.genome !== prevProps.genome ||
+            this.props.minimumBases !== prevProps.minimumBases ||
+            this.props.locus !== prevProps.locus ||
+            this.props.reference !== prevProps.reference ||
+            this.props.tracks !== prevProps.tracks
+        ) {
+            igv.removeAllBrowsers();
```
alexcjohnson, line 35: Oh this line is extra in `componentDidUpdate` - is this really what we want though? What if there are two igv components on the same page? Can we remove just this browser?

src/lib/components/Igv.react.js
```diff
@@ -0,0 +1,79 @@
...
+    String identifier defining genome (e.g. "hg19"). See https://github.com/igvteam/igv.js/wiki/Reference-Genome
+    for details and list of supported identifiers. Note: One (but only one) of
+    either genome or reference properties must be set.
+    */
+
+    genome: PropTypes.string,
+
+    /**
+    Object defining reference genome. see https://github.com/igvteam/igv.js/wiki/Reference-Genome
+    Note: One (but only one) of either genome or reference properties must be set.
```
alexcjohnson, line 56: What happens if you set both? Is an error thrown? Or is one of them ignored?

If it's an error, I'd vote that we avoid it and instead just pick one to omit.

And then once we ignore one of them when both are present, let's document that behavior here.

tests/dashbio_demos/dash-igv/app.py
```diff
@@ -0,0 +1,162 @@
...
+                    dcc.Tab(
+                        label='About',
+                        value='what-is',
+                        children=html.Div(className='control-tab', children=[
+                            html.H4(className='what-is', children='What is IGV?'),
+                            html.P(
+                                """
+                                The Dash IGV component is a high-performance genomics
+                                data visualization component developed originally by the [IGV
+                                Team]( based at UC San Diego and the Broad Institute. It offers
```
alexcjohnson, line 61: This looks like it's trying to be markdown, but it's missing the link and closing paren... could turn this block into `dcc.Markdown`, that would also make the link in the paragraph below active.

tests/integration/test_igv.py
```diff
@@ -0,0 +1,167 @@
...
+
+def test_dbigv001_ASM985889v3(dash_duo):
+    app = dash.Dash(__name__)
+
+    try:
+        requests.get('https://www.google.com/').status_code
+        data_path = 'https://s3.amazonaws.com/igv.org.genomes/'
+    except requests.exceptions.ConnectionError:
+        print("Running test with local datasets")
+        data_path = app.get_asset_url('')
```
alexcjohnson, line 26: Is there a reason not to always use the local datasets in the test?

Also, nonblocking but looks like it would be a little simpler to include the subdiretory and base filename in `data_path` - so below you'd just have `data_path + ".fna"` for example.

HammadTheOne: I think it's important to use an online dataset for at least one of the tests since it seems the more common use-case for  this component would be to load a remotely hosted database. It might be worthwhile to make sure that functionality doesn't break, especially since the `genome` prop pulls data from a list of pre-defined reference genomes hosted by igv. 

On a similar note, based on a [comment by the developer of igv.js here](https://bioinformatics.stackexchange.com/a/10666), and the way local file loading is setup in their web app, it does seem like Dash simplifies loading local files significantly.

package.json
HammadTheOne: Thanks for pointing it out, fixed in dff2f6c.

tests/integration/test_igv.py
alexcjohnson: OK. We've had issues in the past where some external resource went down and broke our CI tests, but I suppose `s3` going down would take a large fraction of the internet with it, so probably the least of our worries 😅

tests/dashbio_demos/dash-igv/app.py
HammadTheOne: Fixed in 0e650d0.

src/lib/fragments/Igv.react.js
HammadTheOne: Fixed in 8db8ccb so that only the browser in question is removed. I hadn't considered that someone might use multiple independent IGV browsers in a single app/page, but it could be a niche use, and it makes for more specific code.

src/lib/fragments/Igv.react.js
HammadTheOne: Changed in b0561a2.

src/lib/fragments/Igv.react.js
HammadTheOne: Changed in 0e650d0.

src/lib/components/Igv.react.js
HammadTheOne: If both are set, the `genome` key is omitted. I've updated the documentation with that behaviour.

alexcjohnson (approved):
Beautiful! 💃 once tests pass (looks like just a minor lint issue)

PR #544, opened 2021-02-15
0.6.1 Patch for NGL fixes

HammadTheOne:

## Description of changes

As part of plotly/dash-docs#1090, this PR adds some minor fixes and maintenance for the NglMoleculeviewer component. The main changes are using `React.ref` and checking for `height` and `width` prop changes on update.


src/lib/components/NglMoleculeViewer.react.js
```diff
@@ -21,9 +21,9 @@ export default class NglMoleculeViewer extends Component {
     }
 }
 
-const defaultHeight = '500px';
+const defaultHeight = '600px';
 
-const defaultWidth = '500px';
+const defaultWidth = '600px';
```
alexcjohnson, line 26: Is it important to increase the default height and width? This has been released, so ideally we don't change its defaults, but if this is substantially better it's OK - we're still on 0.x and it hasn't been out long anyway.

alexcjohnson (approved):
Very nice finds re: both `ref` and height/width updates! Well done. Just one comment to consider re: default changes, then 💃

src/lib/components/NglMoleculeViewer.react.js
HammadTheOne: It's a minor change, but I think it's nice to have. The additional space by default gives a little bit more room to work with when moving the molecule around.

PR #547, opened 2021-03-11
Ngl SideBySide fix and Mol3d isosurface/shapes

### LINKED ISSUE #518: Update Contributing.md and Demos to match current repo structure
Some parts of the `CONTRIBUTING.md` require updates to bring it in line with the currently applied process in the `dash-bio` repo. Alternatively, the directories in the repo that don't comply with the contributing guidelines can be updated to match best practices. For example the following instructions specify snake_case, but the sub-directories' names are hyphenated and include an `app.py`. 

```
In the `tests/dashbio_demos/` subfolder, please create a file named
`app_{your component name in snake case}.py`. In this file, please include the
```
Similarly, the following subfolder and (I believe) the gallery structure no longer exist. 

```
In the `tests/dashbio_demos/images/` subfolder, please include a PNG file named
`pic_{your component name in snake case}.png`.
```

### LINKED ISSUE #526: Molecule3D shapes and isosurfaces
I'm enjoying using the 3D viewer in dash-bio, but I have one question.

Seeming as the 3D viewer is taken from the react wrapper of 3dmol.js, I was wondering why the `shapes` and `isosurface` features aren't implemented? It would be **really** nice to have these available to use in Dash.

Cheers!

HammadTheOne:
Closes #526 
Closes #518 

## About
- [x] I am closing an issue

## Description of changes

This PR adds the isosurfaces and shapes props to the `Molecule3dViewer` component, and fixes an issue with `sideByside` not working correctly with ALL chains selected in `NglMoleculeViewer`. 

- [x] Fix ngl sidebyside fb4ffa1
- [x] Add isosurfaces and shapes c77d79b
- [x] Fix integration tests and Percy snapshots flakiness

TODO: 

- [x] Add new feature documentation to `dash-docs` (https://github.com/plotly/dash-docs/pull/1107)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.6.2] - 2021-03-10
+### Fixed
+* [#545](https://github.com/plotly/dash-bio/pull/545) `sideByside` prop fix for NglMoleculeViewer when `ALL` chains are visualized.
+### Added
+* [#546](https://github.com/plotly/dash-bio/pull/546) Added shapes and isosurfaces props to 3dMoleculeViewer to enable rendering additional features on the molecule.
```
alexcjohnson, line 7: These numbers aren't correct - both of these are in this PR, right?

BTW just a followup on our slack discussion: typically when we're adding features we'll just add an `## Unreleased` heading, with no date or release number, and then replace it with the actual release number and date during the actual release process.

HammadTheOne: Thanks for pointing it out, those were WIP numbers and I've fixed them in 6e5c669, along with the version.

src/lib/components/Molecule3dViewer.js
```diff
@@ -85,10 +85,29 @@ Molecule3dViewer.propTypes = {
     selectedAtomIds: PropTypes.array,
 
     /**
-     * labels corresponding to the atoms of the molecule
+     * Labels corresponding to the atoms of the molecule.
+     * The text key sets the label content, and additional
+     * styling options can be set with the parameters key.
```
alexcjohnson, line 90: Can we document this as part of the propType? So instead of `labels: PropTypes.array` we would have something like (if I'm understanding this correctly):

```js
/**
 * Labels corresponding to the atoms of the molecule.
 */
labels: PropTypes.arrayOf(PropTypes.exact({
    /**
     * The label content
     */
    text: PropTypes.string,

    /**
     * additional styling options. See <docs url> for details
     */
    parameters: PropTypes.object
}))
```
Notice I used `exact` - this is preferred over `shape` unless it's specifically desirable to allow arbitrary extra keys in the object. Should we change this for `orbital` below, or does it support extra keys?

src/lib/fragments/NglMoleculeViewer.react.js
```diff
@@ -202,7 +202,24 @@ export default class NglMoleculeViewer extends Component {
...
+                const strucCenter = struc.getCenter();
+                const pa = structure.getPrincipalAxes();
+
+                struc.setRotation(pa.getRotationQuaternion());
+                struc.setPosition([
+                    0 - strucCenter.x - xOffset,
+                    0 - strucCenter.y,
+                    0 - strucCenter.z,
+                ]);
+                this.addMolStyle(struc, sele, chosen, color);
```
alexcjohnson, line 219: 🌴  duplicated below

src/lib/components/Molecule3dViewer.js
HammadTheOne: For orbital, while it doesn't support extra keys, the component itself supplies defaults for missing keys. I think changing it to `exact` is fine, if the warning for additional or missing keys is what we're looking to have. Fixed in 078ebc7.

src/lib/fragments/NglMoleculeViewer.react.js
HammadTheOne: Fixed in 238f801.

src/lib/components/Molecule3dViewer.js
alexcjohnson: > `exact` is fine, if the warning for additional or missing keys is what we're looking to have

`exact` doesn't warn on missing keys - that's still fine unless the individual key has `.isRequired`. All it does is forbid including additional keys. So yes, this is what we want.

HammadTheOne: After looking at it again, the `parameters` are just input as extra keys to the `label` object. In 3010db8 I changed it to remove the explicit `parameters` prop and changed it back to `shape` so arbitrary keys can be passed in without warnings. Does that make sense for this prop?

src/lib/components/Molecule3dViewer.js
```diff
@@ -88,16 +88,14 @@ Molecule3dViewer.propTypes = {
...
-        PropTypes.exact({
+        PropTypes.shape({
             /**
              * The label content
              */
             text: PropTypes.string,
-
             /**
-             * Additional styling options. See https://dash.plotly.com/dash-bio/molecule3dviewer for details
+             * For additional styling keys, see https://3dmol.csb.pitt.edu/doc/types.html#LabelSpec
```
alexcjohnson, line 97: OIC, it's not `text` and `parameters`, it's `text` and 17 other possible keys to specify the other parameters. I misunderstood that when I first looked at this PR. A comment here (with no following key) won't make it into the auto docs (as you can see from the built Py and R artifacts above, and the docs you DO get in this case are going to look funny if you specify `text` explicitly without specifying all the other ones (which I don't think we should do) - you can add a note about "see <docs> for all the other keys you can include" but it'll still say "labels has the following type: list of dicts containing keys 'text'." which would confuse people about the other keys.

So I guess really the best option here would be to go back toward what you originally had, something like:

```js
/**
 * Labels corresponding to the atoms of the molecule.
 * Each label has a `text` field, a string containing the label content,
 * and can have many other styling fields as described in
 * https://3dmol.csb.pitt.edu/doc/types.html#LabelSpec
 */
labels: PropTypes.arrayOf(PropTypes.object))
```

src/lib/fragments/NglMoleculeViewer.react.js
```diff
@@ -184,6 +184,24 @@ export default class NglMoleculeViewer extends Component {
         }
     }
 
+    // helper function to add molecules to the stage side by side
+    addMolSideBySide(stageObj, sele, stage, chosen, color, xOffset) {
```
alexcjohnson, line 188: Hmm, I'm surprised that the existing custom methods haven't showed problems with this already - perhaps it's only strictly necessary when you attach them to the DOM as event handlers, but for safety sake all the methods here that aren't built-in React methods should be explicitly bound to `this` in the `constructor`, like `circos` does for example:

https://github.com/plotly/dash-bio/blob/3010db8dea220daaf3ddaf4d22fa812022a8b4eb/src/lib/fragments/Circos.react.js#L17-L23

src/lib/components/Molecule3dViewer.js
alexcjohnson: See below https://github.com/plotly/dash-bio/pull/547#discussion_r593419924 - FYI all of this would be clearer if there were a test using the `labels` key so we could both be looking at the same behavior 😉

alexcjohnson: Also, if you want a `dash.testing` test to verify that the prop types are all valid, you can start the server using:

`dash_duo.start_server(app, dev_tools_props_check=True)`

and add at the end of the test:

`assert not dash_duo.get_logs()`

src/lib/components/Molecule3dViewer.js
HammadTheOne: Updated in cb6faa7.

src/lib/fragments/NglMoleculeViewer.react.js
HammadTheOne: Fixed in 4c9097c.

src/lib/components/Molecule3dViewer.js
alexcjohnson: thx - and apologies for the whiplash 😅

src/lib/components/Molecule3dViewer.js
HammadTheOne: I added a Percy snapshot test in the last couple of commits. One thing I'm confused about is that the Percy build isn't showing the new screenshot for approval in the build. Does that require manual setup aside from just adding in `    dash_duo.percy_snapshot('test-mol3d_labels')` to the test?

alexcjohnson: oh boy... we have a problem, all the percy snapshots are getting separate builds on their side. Check out https://percy.io/plotly/dash-bio and see all the ones from the same commit. Try using the `percy-finalize` pattern like we use elsewhere, see eg https://github.com/plotly/dash-html-components/commit/72796c48e3dc7bb1999ef47f9adb6d2f81e6e04a

HammadTheOne: OK, after a long series of commits, the tests are finally consolidated into one build with `percy-finalize`, and running successfully on 3.7 after fixing flakiness for a few of them. 

One thing I am noticing for the Percy snapshots is that the molecules don't seem to be rendered for the `mol-3d` and `ngl-moleculeviewer` tests. I tried adding in a delay with a check using `wait_for_text_to_equal` and a dummy div, but it seems to be the same case, despite loading fine locally.

Do you have any suggestions on how to make these snapshots more reliable?

alexcjohnson: Excellent, nice job! The unrendered pieces are probably in canvases (WebGL or otherwise) - you'll need `convert_canvases` in the snapshot call, like here:

https://github.com/plotly/dash/blob/e8ac94919105a91c76a966c21aca2ec7b0297e22/tests/integration/devtools/test_devtools_ui.py#L103

HammadTheOne: Thanks, that worked perfectly, and now all but the Speck ones are generated successfully! I can't readily see what might be causing it since Speck is also a WebGL rendered component, and locally it doesn't seem like there's a problem, but I think it might be worth addressing that in a separate PR and wrapping this one up (I feel like it's gotten a little out of scope already :sweat_smile: ).

HammadTheOne:
@alexcjohnson Tests successfully running again.

alexcjohnson (approved):
Excellent! 💃

PR #551, opened 2021-04-03
fix volcano margins for legend

akmorrow13:
Closes #[issue number; if this PR doesn't address an issue, delete this line]

## About
- [ ] I am closing an issue
- [ ] This is a new component


akmorrow13:
BEFORE:
Screen Shot 2021-04-03 at 9 57 01 AM
AFTER:
Screen Shot 2021-04-03 at 9 58 51 AM

xhlulu (approved):
Looks so much better! 💃

xhlulu:
@alexcjohnson @HammadTheOne Anything thoughts?

HammadTheOne:
> @alexcjohnson @HammadTheOne Any thoughts?

While it does make the margins what we want them, I'm very hesitant to modify the underlying `volcano_plot` component arguments directly with hardcoded values in this PR. We could also just change the styling of the parent component using CSS selectors (eg. `.control-tabs`) to make it more readable. 

I think these arguments should be allowed to be passed in by users anyways, so I can open a separate PR that enables that and sets the current ones as defaults. I'll link it to this PR once it's opened.

HammadTheOne:
PR for VolcanoPlot updates is opened here: https://github.com/plotly/dash-bio/pull/554

Once that's merged, we'll be able to pass in arguments for the `Layout` of the figure to set margin, height, width, or any other key referenced [here](https://plotly.com/python-api-reference/generated/plotly.graph_objects.Layout.html).

akmorrow13:
Closing in favor of https://github.com/plotly/dash-bio/pull/554

PR #552, opened 2021-04-04
fixed exact error for track props

akmorrow13:
Closes #[issue number; if this PR doesn't address an issue, delete this line]

## About
- [ ] I am closing an issue
- [ ] This is a new component


HammadTheOne:
We'll want to use `exact` instead of `shape` here, unless there's any additional arbitrary keys supported by this prop (which I don't believe there are). That way, we'll have explicit warnings if any extra properties are passed.

I think this PR can be closed without merging, but feel free to re-open it if you have any other thoughts on this.

PR #554, opened 2021-04-07
Update VolcanoPlot props

HammadTheOne:
## About

## Description of changes

As a continuation of #551 , this PR adds the `height/width`, `margin` and `legend` props to the `VolcanoPlot` component. In addition, we add `kwargs` to pass arbitrary values in to the `plotly.graph_objects.Layout` of `VolcanoPlot` so that any arguments available for `Layout` are able to be passed here as well (eg. `bgcolor` or `template`).


akmorrow13:
@HammadTheOne I confirmed this works by setting:

```
dash_bio.VolcanoPlot(
        dataframe=DE_dataframe,
        margin=go.layout.Margin(l = 0,r = 0, b = 0),
        legend= {
            'orientation': 'h',
            'yanchor': 'bottom',
            'y' : 1.02,
            'bgcolor': '#f2f5fa'
        }
)
```

Should we set the defaults to something that looks cleaner?

HammadTheOne:
> @HammadTheOne I confirmed this works by setting:
> 
> ```
> dash_bio.VolcanoPlot(
>         dataframe=DE_dataframe,
>         margin=go.layout.Margin(l = 0,r = 0, b = 0),
>         legend= {
>             'orientation': 'h',
>             'yanchor': 'bottom',
>             'y' : 1.02,
>             'bgcolor': '#f2f5fa'
>         }
> )
> ```
> 
> Should we set the defaults to something that looks cleaner?

Thanks for confirming that it worked! I think having a nice, clean default legend/layout is a good idea. We had this originally, and I think we should have a non-white bgcolor and have the legend outside the plot area:

```
legend={
    'x': 0.85,
    'y': 0.1,
    'bgcolor': '#f2f5fa'
}
```

I'll make an update for those, but if you have any suggestions, I'd love to hear them. We can use the [Plotly graphing library reference](https://plotly.com/python/legend/#styling-legends) for inspiration.

dash_bio/component_factory/_volcano.py
```diff
@@ -322,6 +368,10 @@ def figure(
     - highlight_color (string; default 'red'): Color of the data points
         highlighted because considered significant. Can be in any color
         format accepted by plotly.graph_objects.
+    - Additional keys (misc.): Arbitrary arguments can be passed to modify the
+        Layout and styling of the graph. A full reference of acceptable args is
+        available [here](https://plotly.com/python-api-reference/generated/plotly.graph_objects
+        .Layout.html).
```
alexcjohnson, line 374: Since we're accepting `**kwargs` here and passing them to `go.Layout`, do we actually need _any_ of the other args you're adding here? We can list some of the common ones (including the ones you're including explicitly now) in the description here, they'd all work identically as part of kwargs.

`x_axis` and `y_axis`: in the docstring there's no `_` and I'd argue it's cleaner without, then these can just be left to kwargs as well.

In order to keep the defaults you have but allow users to update the layout arbitrarily with kwargs after that, I suspect the easiest would be to generate the default layout first, then do `layout.update(**kwargs)`. A little exploring shows that this does a deep merge, so if your kwargs are just `xaxis=dict(color="red")` you'll still get the rest of the x axis as set in the default layout.

HammadTheOne: > do we actually need any of the other args you're adding here?

I thought it might be nice to have the most commonly used args available explicitly, especially for users that use IDE's which show parameter info, but simply including them as part of the documentation is also an option. However, retaining them as explicit args would also potentially save users a trip to the Plotly graphing library docs. I think it's a trade-off between cleaner params (with everything as kwargs) and easier "default" use of the component.

alexcjohnson: Fair point re: IDE docs & completion. That said for everything except `height` and `width` you're still going to need to visit the graphing library docs to find out what goes inside the object, and it feels a bit arbitrary to be pulling in just a few layout options - as if the ones in kwargs are less-fully-supported or something. So I still think we'll be better off removing all of them and just documenting a few (the set you've proposed adding here looks good) in the docstring.

HammadTheOne: > you're still going to need to visit the graphing library docs to find out what goes inside the object,

That's a great point - Updated in a5b996a.

HammadTheOne:
@alexcjohnson  I think the last couple of commits should address the rest of the points you brought up, and the tests are now passing successfully.

dash_bio/component_factory/_volcano.py
```diff
@@ -75,10 +76,21 @@ def VolcanoPlot(
 - title (string; default 'Volcano Plot'): Title of the graph.
 - xlabel (string; optional): Label of the x axis.
 - ylabel (string; default '-log10(p)'): Label of the y axis.
+- xaxis (dict: optional): Dict with compatible properties for the x-axis of
+     the figure layout.
+- yaxis (dict: optional): Dict with compatible properties for the y-axis of
+    the figure layout.
```
alexcjohnson, line 82: Let's move these (and height, width, margin, legend below) into the "Additional keys" section of the docstring, saying something like "Some commonly-used keys are:" and then exactly the text you have currently. These keys don't explicitly show up in the function signature so it'll be confusing to have them mixed in like this.

dash_bio/component_factory/_volcano.py
```diff
@@ -274,9 +290,11 @@ def __init__(
 
     def figure(
             self,
-            title='Volcano Plot',
+            title=None,
             xlabel=None,
             ylabel='-log10(p)',
+            xaxis=None,
+            yaxis=None,
```
alexcjohnson, line 297: Is there any difference between explicitly including these to pass to `figure.update_xaxes(**xaxis)`, and omitting them so they get picked up by `layout.update(**kwargs)`? I suppose if we were to have multiple x or y axes, the former would update all of them, the latter would only update the first pair. But I don't believe we do have multiple axes, and anyway if we did I imagine you'd want the ability to update them separately.

dash_bio/component_factory/_volcano.py
```diff
@@ -355,13 +389,18 @@ def figure(
             col = 'black'
 
         layout = go.Layout(
-            title=title,
+            title={'text': 'Volcano Plot',
```
alexcjohnson, line 392: Should we be using the `title` arg here? As it is we're ignoring it, right? Though passing the `title` arg directly to `title.text` in the layout means you can't use the dict form of `title` to modify its styling. Then the only way to do that would be to use magic underscores in the kwargs (`title_font_color="red"` etc).

Alternatively we could drop `title` as an arg completely, so it gets picked up by `layout.update`. The downside of that approach is that passing in a string for `title` would wipe out the default style you have below - at least until https://github.com/plotly/plotly.py/issues/3148 gets fixed, which hopefully will not be too long (🙏  @nicolaskruchten )

So yeah, I guess my inclination is to drop `title` as an arg, even though it's a breaking change in the short term, because the current behavior will be restored with much more flexibility after we fix https://github.com/plotly/plotly.py/issues/3148

dash_bio/component_factory/_volcano.py
HammadTheOne: Updated in e3dfcb4.

dash_bio/component_factory/_volcano.py
HammadTheOne: I was worried that key-worded arguments wouldn't be unpacked down to `xaxes` without overwriting the defaults, but after seeing your comment about the deep merge and testing it out, there is no difference. Fixed in 56948c3.

dash_bio/component_factory/_volcano.py
HammadTheOne: Dropped in 3e2ce66.

alexcjohnson (approved):
Looks great! The change is pretty simple now, which bodes well :) 💃

PR #557, opened 2021-04-19
0.7.0 Release

HammadTheOne:
This PR includes some final updates to make the dash-bio-0.7.0 release. Includes version and changelog updates, and one final run of CI tests to make sure there's no issues.


HammadTheOne:
@alexcjohnson I think we're ready for the 0.7.0 release. If there's nothing else here that stands out, I can go ahead and make the release to `npm` and PyPi.

alexcjohnson (approved):
LGTM. The only other PRs we have open right now that could be considered are the dep updates, but one isn't a big deal and the other is too big a deal so broken. Should do that at some point but doesn't need to be now. 💃

PR #562, opened 2021-06-08
Fix 3dmol zoom and expose zoom props

### LINKED ISSUE #510: Repeated zooming after updating a Molecule3dViewer
**Describe the bug**
I have a small dash app that includes a 2D graph and a Molecule3dViewer, such that when I click on points in the 2D graph I update the Molecule3dViewer. My dash callback returns something like this, following the example here.

```
    return dash_bio.Molecule3dViewer(
            id='mol-3d',
            selectionType='atom',
            modelData=mdata,
            styles=data_style,
            selectedAtomIds=[],
            backgroundOpacity='0',
            atomLabelsShown=False,
        )
```
I noticed that when I repeatedly click on the graph, each time the Molecule3dViewer updates, the molecule shrinks slightly.

**Expected behavior**
My expectation would be that updating the Molecule3dViewer multiple times would not repeatedly shrink the displayed molecule.

I believe that the issue is that this line of code executes every time the Molecule3dViewer is updated, causing the effective zoom to get decreased multiplicatively. I presume the intent was for this to only happen once. https://github.com/plotly/dash-bio/blob/master/src/lib/fragments/Molecule3dViewer.js#L18

I installed a modified version of the code on my machine where I simply replaced that number with 1.0 by manually editing the compressed javascript bundle, and it worked, which I think confirms my diagnosis.

HammadTheOne:
Closes #510 

## About
- [x] I am closing an issue

## Description of changes

Repeated re-rendering of the Molecule3dViewer component caused the viewer zoom to decrease multiplicatively.
As part of this fix, we also expose the `zoom` and `zoomTo` props for this component to be available for Dash Callbacks.


src/lib/fragments/Molecule3dViewer.js
```diff
@@ -14,9 +14,16 @@ export default class Molecule3dViewer extends Component {
...
+        glviewer.zoomTo(
+            this.props.zoomTo.sel,
+            this.props.zoomTo.animationDuration,
+            this.props.zoomTo.fixedPath
+        );
+        glviewer.zoom(
+            this.props.zoom.factor,
+            this.props.zoom.animationDuration,
+            this.props.zoom.fixedPath
+        );
```
alexcjohnson, line 26: This looks great for initial render, but I'm trying to figure out what will happen on various kinds of updates.

According to https://www.npmjs.com/package/molecule-3d-for-react:

> onRenderNewData {function} [(glviewer) => {}]
> A callback for when the modelData has changed and it has been re-rendered in the viewer.

So we're only going to get here when `modelData` changes, even though many other props may have changed too? Or do we actually get here on any rerender?

Specifically here are two cases I'm interested in:
1. The app has a toggle to turn some features on/off - labels or orbitals perhaps - and the user rotates & zooms (also I just discovered you can pan with ctrl-drag) then toggles those features. Do we snap back to the original view? We shouldn't, we should keep the view they found.
2. The app has a "reset zoom" button, or a button to zoom to a particular view, that just gives new `zoom` and/or `zoomTo` props but leaves all the other props unchanged. Does this work? ie the user zooms, rotates, and pans, does this button get back to the desired state?

src/lib/components/Molecule3dViewer.js
```diff
@@ -25,6 +25,16 @@ Molecule3dViewer.defaultProps = {
...
+    zoom: {
+        factor: 0.8,
+        animationDuration: 0,
+        fixedPath: true,
+    },
+    zoomTo: {
+        sel: {},
+        animationDuration: 0,
+        fixedPath: true,
+    },
```
alexcjohnson, line 37: Nested values behave oddly in `defaultProps` - if you just provide `zoom: {factor: 2}` you'll lose `animationDuration: 0` and `fixedPath: true`. That's fine for `animationDuration` because it seems like its default is in fact `0`, but `fixedPath` looks like it defaults to `false`? If I'm understanding it correctly though, maybe we'd actually be better off with `false`? The [docs](http://3dmol.csb.pitt.edu/doc/$3Dmol.GLViewer.html#zoom) say:

> if true animation is constrained to requested motion, overriding updates that happen during the animation

In the `zoomTo` / `zoom` chain we have, is `zoom` an "update that happens during the animation" of `zoomTo`? If I wanted a button that animates to a particular view, from whatever the user had focused on, would this make a smooth single 2-second animation to that view?
```js
zoomTo: {sel: importantThing, animationDuration: 2000, fixedPath: false}
```
(noting that `onRenderNewData` will call `glviewer.zoom` after `glviewer.zoomTo` no matter what, but I'm happy with the default 0.8 zoom as an endpoint)

src/lib/components/Molecule3dViewer.js
```diff
@@ -118,6 +128,48 @@ Molecule3dViewer.propTypes = {
...
+     */
+    zoom: PropTypes.exact({
+        /**
+         * Magnification factor. Values greater than 1 will zoom,
+         * in, less than one will zoom out. Default 2.
+         */
+        factor: PropTypes.number,
+        /**
+         * An optional parameter that denotes the duration of a
+         * zoom animation.
```
alexcjohnson, line 143:
```suggestion
         * zoom animation, in milliseconds.
```
Never stated in their docs but must be ms? Same for `zoomTo`.

src/lib/fragments/Molecule3dViewer.js
HammadTheOne: I think you've raised a good point here, and this might be require changes to the behaviour of the component when it updates. So far, I think the way we've handled updates to props with Molecule3dViewer is by re-rendering the container and updating the parent either `div` or `dcc.Loading` with our callbacks. 

I'm noticing that if we do attempt to update props directly with a callback, the component doesn't re-render as it should, and the loading state is stuck.

The following snippet shows this when the button is clicked to change the `backgroundColor` prop - the prop changes, but the loading state remains `true`. 

```
import json
import six.moves.urllib.request as urlreq
from six import PY3

import dash
import dash_bio as dashbio
import dash_html_components as html
import dash_core_components as dcc

app = dash.Dash(__name__)

model_data = urlreq.urlopen(
    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/' +
    'mol3d/model_data.js'
).read()
styles_data = urlreq.urlopen(
    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/' +
    'mol3d/styles_data.js'
).read()

if PY3:
    model_data = model_data.decode('utf-8')
    styles_data = styles_data.decode('utf-8')

model_data = json.loads(model_data)
styles_data = json.loads(styles_data)

app.layout = html.Div([
    dashbio.Molecule3dViewer(
        id='my-dashbio-molecule3d',
        styles=styles_data,
        modelData=model_data
    ),
    html.Button(id="button", n_clicks=0, children="Toggle background")
])


@app.callback(
    dash.dependencies.Output('my-dashbio-molecule3d', 'backgroundColor'),
    [dash.dependencies.Input('button', 'n_clicks')],
    prevent_initial_call=True
)
def toggle_labels(n_clicks):
    if n_clicks > 0:
        return("#FF0000")


if __name__ == '__main__':
    app.run_server(debug=True)
```

It doesn't look like our `render()` method would need any other data, so I'm a little confused why it wouldn't re-render by default.

src/lib/components/Molecule3dViewer.js
HammadTheOne: :+1:  Fixed in fe76521.

src/lib/components/Molecule3dViewer.js
HammadTheOne: The main difference in practice is that `fixedPath` forcibly maintains the view as it zooms in - rotating the view snaps it back to the original view it started at while zooming in. I agree that it makes sense to set it to `false` by default, it's less intrusive and more smooth. This does mean that you couple the two together - the animation view begins from whatever you set the `zoom` to rather than some predefined default. 

As far as chaining animations, that is correct, it would make a single 2-second animation which begins from and maintains the specified `zoom` factor to the `sel` residue. 

 On a side note, there's pretty limited documentation on the `zoomTo` selection syntax, but adapting over [the 3dMol.js tests](https://github.com/3dmol/3Dmol.js/blob/bb653fa0cc9d940b78d5c61d7a7390ff3dab471c/tests/auto/tests/test98.js) worked out! I've added additional `PropType` constraints which should help make this a bit clearer. Here's the snippet for a small demo.
 
 I've made the changes in c82d1a9.
 
 ```
 import json
import six.moves.urllib.request as urlreq
from six import PY3

import dash
import dash_bio as dashbio
import dash_html_components as html
import dash_core_components as dcc

app = dash.Dash(__name__)

model_data = urlreq.urlopen(
    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/' +
    'mol3d/model_data.js'
).read()
styles_data = urlreq.urlopen(
    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/' +
    'mol3d/styles_data.js'
).read()

if PY3:
    model_data = model_data.decode('utf-8')
    styles_data = styles_data.decode('utf-8')

model_data = json.loads(model_data)
styles_data = json.loads(styles_data)

app.layout = html.Div([
    dashbio.Molecule3dViewer(
        id='my-dashbio-molecule3d',
        styles=styles_data,
        modelData=model_data,
        zoomTo={
            "sel": {"chain":'X',"resi":115},
            "animationDuration": 2000,
            "fixedPath": False
        },
        zoom={"factor": 0.6}
    ),
])




if __name__ == '__main__':
    app.run_server(debug=True)
```

src/lib/fragments/Molecule3dViewer.js
HammadTheOne: Sorry, false alarm, this doesn't seem to be the case, it responds correctly to re-renders, the component was just missing a condition in `shouldComponentUpdate` for `labels`. For the two app scenarios, here's what we have:

1. If we rotate/zoom, and then update or toggle the props (eg.`backgroundColor` or `labels`), the present view remains the same

2. This one doesn't work. I think, as you pointed out, the `glviewer` view will only change on initial render. I'll try to find a workaround for this, but it might involve adding a method to the Molecule3DViewer component itself.

alexcjohnson: You should be able to do it without modifying `Molecule3d`
- store `glviewer` in the instance so you have it available (`this.glviewer = glviewer` during `onRenderNewData`)
- store the `zoom` and `zoomTo` props similarly after every `glviewer.zoomTo/zoom` chain.
- add a `componentDidUpdate` method that checks if `zoom` and `zoomTo` are `===` their previous values; if they've changed (and you have a `glviewer`), `this.glviewer.zoomTo/zoom` and store the props again.

HammadTheOne: Thanks for the suggestion, added in aa7bbfe.

One question - Is there any particular reason we need to store the `zoom` or `zoomTo` values? We can access them with `prevProps` in `componentDidUpdate` for the same functionality so I'm confused why we'd need to store them again.

alexcjohnson: Oh I forgot about that - good call!

alexcjohnson (approved):
💃 LGTM!

PR #563, opened 2021-06-17
Clean up build scripts and imports

### LINKED ISSUE #534: use _imports_, not load_components
`__init__.py` is still using `_dash.development.component_loader.load_components` - this has long been obsoleted by the build process creating `_imports_.py`. See eg [dcc's `__init__.py`](https://github.com/plotly/dash-core-components/blob/dev/dash_core_components_base/__init__.py)

### LINKED ISSUE #556: Update package scripts for dash-bio
As per the comment in #548, the package script for dash-bio should be updated to reflect the current repo structure and bring them in line with other component libraries:

> using perhaps dcc as a model: there should only be three build scripts: build:js, build:backends (builds Py, R, and Jl using dash-generate-components, no python -c, no get_version_info.py), and build to combine them. Assuming our source mapping is set up correctly there is no reason for build:dev or build:js-dev.

This will clean up unnecessary scripts and make it easier to develop and track the steps during package build.

HammadTheOne:
Closes #534 
Closes #556 

## About
- [x] I am closing an issue

## Description of changes

This PR changes `__init__.py` to use `__imports__` generated by the build process instead of the `_dash.development.component_loader.load_components` method. It also cleans up the `package.json` in line with what we have for `dcc`, with 3 standard build scripts for the `js` and `backends`.


dash_bio/__init__.py
```diff
@@ -115,7 +110,6 @@
 _css_dist = []
 
 
-for _component in _components:
-    setattr(_this_module, _component.__name__, _component)
-    setattr(_component, '_js_dist', _js_dist)
-    setattr(_component, '_css_dist', _css_dist)
+for _component in __all__:
+    setattr(locals()[_component], "_js_dist", _js_dist)
+    setattr(locals()[_component], "_css_dist", _css_dist)
```
alexcjohnson, line 115: not a big deal, but turns out since there's no separate CSS we can drop `_css_dist` entirely. [dcc](https://github.com/plotly/dash-core-components/blob/72875fc254dcfe461d77ae649effb51a0ecb16c9/dash_core_components_base/__init__.py#L142-L143) works this way for example.

dash_bio/package-info.json
```diff
@@ -18,14 +18,10 @@
...
     "test-unit": "jest",
-    "build:js-dev": "webpack --mode development",
     "build:js": "npm run lint && npm run format && webpack --mode production",
-    "build:py": "node ./extract-meta src/lib/components > dash_bio/metadata.json && python get_version_info.py && python -c \"import dash; dash.development.component_loader.generate_classes('dash_bio', 'dash_bio/metadata.json')\"",
-    "build:r": "dash-generate-components ./src/lib/components dash_bio -p package-info.json --r-prefix='dashbio' --r-imports='dash' --r-depends='manhattanly'",
-    "build": "npm run build:js && npm run build:py",
-    "build:all": "npm run build && npm run build:r",
-    "postbuild": "es-check es5 dash_bio/*.js",
-    "build:dev": "npm run build:js-dev && npm run build:py"
+    "build:backends": "node ./extract-meta src/lib/components > dash_bio/metadata.json && python get_version_info.py && dash-generate-components ./src/lib/components dash_bio -p package-info.json --r-prefix='dashbio' --r-imports='dash' --r-depends='manhattanly'",
```
alexcjohnson, line 22: `extract-meta` is now part of [dash-generate-components](https://github.com/plotly/dash/blob/efc42b46fef033bd4ec1f082cfe5cbab97a62646/dash/development/component_generator.py#L58-L77) so I think we can remove that (and in fact remove that file from this repo 🎉 )

I believe the same is true for `get_version_info` - if you look at the `package-info.json` that's present in the `dash-bio` dir, that's the full `package.json` as copied by `dash-generate-components`, not the partial one from `get_version_info`.

dash_bio/package-info.json
```diff
@@ -18,14 +18,10 @@
...
+    "build": "npm run build:js && npm run build:backends",
+    "postbuild": "es-check es5 dash_bio/*.js"
   },
   "author": "The Plotly Team <dashbio@plotly.com>",
   "maintainer": "Ryan Patrick Kyle <ryan@plotly.com>",
```
alexcjohnson, line 27: We should probably change this... yourself?

HammadTheOne: Yep, changed in 4e6cf1d.

dash_bio/package-info.json
HammadTheOne: That makes a lot of sense - it's nice that `dash-generate-component` bundles a lot of these parts of the build process together; I wasn't entirely sure which of these commands were deprecated. 

Updated in 4e6cf1d.

dash_bio/__init__.py
HammadTheOne: Dropped in 84ffb03   :scissors:

alexcjohnson (approved):
💃

dash_bio/__init__.py
```diff
@@ -115,7 +110,5 @@
 _css_dist = []
```
alexcjohnson, line 110: This line can be dropped as well now... not a big deal though :)

HammadTheOne: Thanks for the catch! 

:hocho:  in 02144f9

PR #564, opened 2021-06-23
Add support for Julia components to be generated

HammadTheOne:
## About

## Description of changes
This PR fixes an incompatibility with `dash-generate-components` and the way some of the `propTypees` were referenced which prevented Julia versions of all components from being generated.

In addition, the Julia prefix will be added to the build script as part of https://github.com/plotly/dash-bio/pull/563 so that all back-ends can be generated at the same time.


alexcjohnson:
Thanks @HammadTheOne - so to be clear the issue is that we were referencing a variable within propTypes, and the value needs to be inlined. Can you open an issue in dash to have the component generator throw an error when this happens? The Python and R components did generate, but were missing the values here prior to this fix, and we'd be better off pushing component authors to fix this kind of problem up front rather than expecting them to notice a problem later and fix it.

Also: The Julia package is being built in a separate repo https://github.com/plotly/DashBio.jl so it seems like we shouldn't include Julia artifacts here.

alexcjohnson (approved):
Excellent! 💃

PR #570, opened 2021-08-31
Fixing CI config for Rust install in `dash`

HammadTheOne:
This PR adds some minor fixes to the CI config in relation to the addition of the `orjson` dependency in `dash-2.0.0`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,12 +54,12 @@ jobs:
             - run:
                 name: Install Dash for package build and build package
                 command: |
-                    pip install --upgrade pip
                     curl --proto '=https' --tlsv1.2 -sSf -o rust-init.sh https://sh.rustup.rs
                     chmod ugo+x rust-init.sh
                     ./rust-init.sh -y
                     python -m venv venv
                     . venv/bin/activate
+                    pip install --upgrade pip
                     git clone --depth 1 https://github.com/plotly/dash.git dash-main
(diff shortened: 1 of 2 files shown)


alexcjohnson:
I still don't get this rust issue, feels like a red herring to me, and I don't want us to get in the habit of installing it everywhere "just in case". What are you seeing that makes this seem necessary? The `dash` and `plotly.py` tests run just fine, both install `orjson` in the same kind of environment, and both should be implicitly using it, just as we do here and in dashR.

cc @homer6

HammadTheOne:
> I still don't get this rust issue, feels like a red herring to me, and I don't want us to get in the habit of installing it everywhere "just in case". What are you seeing that makes this seem necessary? The dash and plotly.py tests run just fine, both install orjson in the same kind of environment, and both should be implicitly using it, just as we do here and in dashR.

Seems like just upgrading pip is an easier fix.

.circleci/config.yml
```diff
@@ -108,6 +111,7 @@ jobs:
                 name: Install Dash
                 command: |
                     . venv/bin/activate
+                    pip install --upgrade pip
```
alexcjohnson, line 114: Here you're in the same venv as [line 102](https://github.com/plotly/dash-bio/pull/570/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R102) so this should be redundant

alexcjohnson (approved):
💃

PR #571, opened 2021-09-04
Fix eventDatum warning in circos component

### LINKED ISSUE #501: Warning - eventDatum prop on a DOM element
Hello,
I keep getting the following warning in Chrome debug console.   It is just a warning and as far as I can tell it is harmless, but the stack trace is >100 lines, so it is a bit of a nuisance to debug with..

```Warning: React does not recognize the `eventDatum` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `eventdatum` instead. If you accidentally passed it from a parent component, remove it from the DOM element.```

It is pretty easy to reproduce using the example on [Circos Examples and Reference](https://dash.plotly.com/dash-bio/circos) page.

Using Python 3.7.7, dash-bio 0.4.8, MacOS Catalina, Chrome 81.0.4044.92

Thanks.

### PR DESCRIPTION
**Pull request by nickmelnikov82**
Closes #501

## About
- [ ] I am closing an issue
- [ ] This is a new component

## Description of changes

I've checked the Circos demo app, eventDatum updating working correctly after these changes

### CODE REVIEW

File: src/lib/fragments/Circos.react.js [hunk 1]
```diff
@@ -264,7 +264,7 @@ export default class Circos extends Component {
         } = this.props;
 
         return (
-            <div id={id} style={style} eventDatum={eventDatum}>
+            <div id={id} style={style} data-event-datum={eventDatum}>
```
HammadTheOne, line 267:
```suggestion
            <div id={id} style={style} eventdatum={eventDatum}>
```
I think we can just set this to `eventdatum` like the warning suggests, for clarity.

HammadTheOne:
I'm noticing 9 failing tests, and these fail locally as well, so it's not just flakiness. This occurs if the dash version is `2.0.0`.

We should investigate this and and update the tests accordingly.

HammadTheOne:
💃 

Great!

HammadTheOne (approved):

PR #574, opened 2021-09-13
Rename requirements.txt into dev-requirements.txt

### LINKED ISSUE #490: Unused root-level requirements.txt file should be replaced with dev-requirements
This was originally used for the demo apps but I don't think it needs to be here any more after https://github.com/plotly/dash-bio/pull/459 since the apps each have their own `requirements.txt`. We should probably replace it with a `dev-requirements.txt` file instead.

_Originally posted by @shammamah in https://github.com/plotly/dash-bio/pull/489_

nickmelnikov82:
Closes #490

## About
- [x] I am closing an issue
- [ ] This is a new component


HammadTheOne (approved):
Looks good to me 💃

PR #575, opened 2021-09-18
Bump the Ideogram version to 1.6.0

### LINKED ISSUE #524: Ideogram annotations do not work or behave erractically.
Okay, I thought this was me not being able to figure out how the data was meant to be formatted. It's not though. 

Annotation tracks are not being loaded for ideograms, either through a URL, or through a local list of dicts with the specified inputs (name, chr, start, stop) listed in the documentation. 
If I load the example provided, using the url for the json in the example code, it works and annotations are loaded, but only for the X and Y chromosome. If I change the specified chromosomes to 1 and 2, the chromosomes display, but annotations do not. If I change only one of the chromosomes, i.e. to ["1","Y"], both chromosomes display, but neither have annotations. There are definitely annotations in the example json file for the specified chromosomes. 

Loading annotations from a local file using annotationsPath does not work. I'm not the only person to [report this](https://github.com/plotly/dash-bio/issues/495). If I download the example file from the url and restrct the ideogram to the two chromosomes, X and Y, that work when loading from a url, then the chromosomes are displayed, but the annotations are not. 

Loading from a list of dicts using the annotations identifier does not work. The documentation suggests that when using this method, one should specify name, chr, start and stop. Creating a list:
'''
annotationList = [{"name":"wibble","chr":1,"start":1,"stop":10000}]
'''
and then calling
'''
  dashbio.Ideogram(
        id='my-dashbio-ideogram',
        annotationsLayout="overlay",
        annotations=annotationList
)
'''
results in the chromosomes being displayed, but again, without annotations. If I switch to annotationsLayout="tracks", again, no annotations are displayed. Switching to annotationsLayout="histogram", annotations are displayed, though if you rotate the chromosome it causes the length of the chromosome to be compressed. 

Additionally, it is not possible to have multiple tracks, as suggested by the documentation. Constructing the following annotation List and track list:
'''
annotationsList = [{"name":"wibble","chr":"1","start":100000,"stop":10000001,"color":"#0F0", "track":"1"}, {"name":"wiaabble","chr":"4","start":100000,"stop":300001,"color":"#0F0", "track":"2"}]
trackList = [{"id":"1","name":"bob","color":"#0F0"},{"id":"2","name":"alic","color":"#00F"}]
'''
and then using those in the layout
'''    dashbio.Ideogram(
        id='my-dashbio-ideogram',
        annotationsLayout="histogram",
        chromosomes=['1', '4'],
        annotationTracks=trackList,
        annotations=annotationsList,
        annotationHeight="1"
    ),
'''
again results in no annotations being displayed. 

In addition to annotations displaying in a single track, as best I can understand from the documentation, I should be able to specify multiple tracks with different colours using annotationTracks. 

This is using Python 3.6.11 in a new conda.

nickmelnikov82:
Addresses some of the issues in #524

## About
- [x] I am closing an issue
- [ ] This is a new component


HammadTheOne (approved):
:woman_dancing:

PR #579, opened 2021-09-28
Add resizing plot feature and created tests for this feature.

### LINKED ISSUE #545: Resizing needle plot
I would like to resize the needle plot so I can make it arbitrarily wide. The default size is a bit smaller than a third of my screen size and considering the legend it is too small to provide a good user experience.

PS: apologies if this is not the right channel to communicate such requirements

nickmelnikov82:
Closes #545 

## About
I am adding a feature to an existing component.

## Description of changes
Create a new feature.
I added height and width properties to the NeedlePlot component and fragment.  In the NeedlePlot component, I add height and width properties as well set them.  In the NeedlePlot component, I gave the div and plot component these properties.
Also, I created tests for checking that these properties work as expected.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,6 +4,7 @@
 cython>=0.19
 dash-bio-utils==0.0.6
 dash-daq==0.5.0
+flake8==3.9.2
 gunicorn==19.9.0
 jsonschema==2.6.0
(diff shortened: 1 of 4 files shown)


HammadTheOne (approved):
:woman_dancing:

PR #591, opened 2021-11-02
Remove height and width default values to make auto-sizing works

nickmelnikov82:
## About
- [ ] I am closing an issue
- [ ] This is a new component


HammadTheOne (approved):
:woman_dancing:

PR #629, opened 2021-11-18
Pass to Igv new array with tracks instead of using directly from props.

### LINKED ISSUE #572: igv duplicate sequence tracks in igv is inside tabs
python: 3.8
dash: 1.20.0
dash-bio: 0.7.1

**Describe the bug**
When dashbio Igv is embeded inside a tab and the user switches between tabs. Coming back to the tab where the igv is triggers a reload of the igv module along with addition of duplicate sequence tracks each time the tab is switched. This behavior seem to be happening when `tracks` are defined. No issue when `tracks` is blank.

This results in the following (note the duplicate sequence tracks in color),
Screen Shot 2021-09-07 at 5 16 31 PM

**To Reproduce**
Below is a toy example to reproduce the behavior. I've replaced urls for the vcf with `LINK`, edit this to refer to an actual file to make the below example work. Same if the tracks were bam alignments, etc.
```
import dash
import dash_core_components as dcc
import dash_html_components as html
import dash_bio as dashbio

app = dash.Dash(__name__)
app.title = 'test'
app.config['suppress_callback_exceptions'] = True
app.layout = html.Div([
    dcc.Tabs(value='tab1', children=[
        dcc.Tab(id='tab1', value='tab1', label='Tab1', children=[]),
        dcc.Tab(id='tab2', value='tab2', label='Tab2', children=[]),
        dcc.Tab(id='tab3', value='tab3', label='Tab3', children=[
            dashbio.Igv(
                id='igv',
                genome='hg38',
                tracks = [
                    {
                        'type': "variant",
                        'format': "vcf",
                        'url': "LINK",
                        'indexURL': "LINK",
                        'name': "VCF",
                    }
                ]
            ),
        ])
    ]),
])

app.run_server()
```

nickmelnikov82:
Closes #572

## About
- [x] I am closing an issue
- [ ] This is a new component

## Description of changes


HammadTheOne (approved):
Tested it out and the fix works great! Adding a changelog entry and merging it.

:woman_dancing:

PR #670, opened 2022-01-27
WIP: Add tests for components

nickmelnikov82:
## Description of changes

Adding tests for all components. Check that each property in each component does what it supposed to do.


HammadTheOne:
This is great so far @nickmelnikov82 and adds some much-needed coverage for our tests. Let me know when it's ready for a final review and I can take a full pass over it.

PR #701, opened 2022-07-14
remove numpy pin in demos

alexcjohnson:
`numpy` has an update that resulted in a million dependabot PRs. We can't accept those because newer `numpy` versions are restricted to Py3.8+ - but we can simply unpin it and folks will get the top version available for their Python version.


alexcjohnson:
Not sure what's up with the speck tests, but that's not a problem introduced by this PR, so I'll merge - just want to get all those dependabot PRs off my queue 😅

